PR #39, opened 2022-04-16
Avoid unnecessary rebuilds

MikailBag:
Crate published on crates.io does not contain ydb-api-protos directory. Hence, `rerun-if-chaned` directive makes cargo rebuild this crate every time. I think (but haven't tested yet) that proposed patch should address this issue.


rekby:
@MikailBag Hello, thanks for fix.

Are you agree with Contributor License Agreement https://yandex.ru/legal/cla/?lang=en

MikailBag:
Yes, I agree with Yandex CLA.

rekby:
Thanks.
Updated crates published as ydb-0.1.1 and ydb-grpc-0.0.7

PR #123, opened 2023-02-12
feat: create wrapper structs for create topic method

AlexKrudu:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

### DIFF
--- a/ydb/src/grpc_wrapper/raw_topic_service/common/codecs.rs
+++ b/ydb/src/grpc_wrapper/raw_topic_service/common/codecs.rs
@@ -1,0 +1,36 @@
+use itertools::Itertools;
+use ydb_grpc::ydb_proto::topic::{Codec, SupportedCodecs};
+
+#[derive(serde::Serialize)]
+pub(crate) enum RawCodec {
+    Unspecified,
+    Raw,
+    Gzip,
+    Lzop,
+    Zstd,
+    Custom(i32),
+}
+
+#[derive(serde::Serialize)]
+pub(crate) struct RawSupportedCodecs {
+    pub codecs: Vec<RawCodec>,
+}
+
+impl From<RawSupportedCodecs> for SupportedCodecs {
+    fn from(value: RawSupportedCodecs) -> Self {
+        Self {
+            codecs: value
+                .codecs
+                .into_iter()
+                .map(|x| match x {
+                    RawCodec::Custom(val) => val,
+                    RawCodec::Unspecified => Codec::Unspecified.into(),
+                    RawCodec::Raw => Codec::Raw.into(),
+                    RawCodec::Gzip => Codec::Gzip.into(),
+                    RawCodec::Lzop => Codec::Lzop.into(),
+                    RawCodec::Zstd => Codec::Zstd.into(),
+                })
+                .collect_vec(),
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


ydb/src/grpc_wrapper/raw_topic_service/common/codecs.rs
```diff
@@ -0,0 +1,26 @@
+use itertools::Itertools;
+use ydb_grpc::ydb_proto::topic::SupportedCodecs;
+
+#[repr(i32)]
+#[derive(serde::Serialize)]
+pub(crate) enum RawCodec {
```
rekby, line 6: 1. Custom is only start point of customer codecs, not fixed value.
The codec value can be 10001 or 12345.

2. When we will read grpc stream it can contain values, not specified by current proto-enumeration and we need to handle it.

What if:
1. Do custom variant with value? Custom(i32) and store all unknown value to the custom field
2. remove repr directive - it will bad for nested values and any way we no need any special layout.

or store RawCodec as struct with only one i32 field + some constants for well known values?

ydb/src/grpc_wrapper/raw_topic_service/common/consumer.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use crate::grpc_wrapper::raw_topic_service::common::codecs::RawSupportedCodecs;
+use ydb_grpc::google_proto_workaround::protobuf::Timestamp;
+use ydb_grpc::ydb_proto::topic::{Consumer, SupportedCodecs};
+
+#[derive(serde::Serialize)]
+pub(crate) struct RawConsumer {
+    pub name: String,
+    pub important: bool,
+    pub read_from: Timestamp,
```
rekby, line 11: Timestamp is protobuf type, it will problem on higher level if grpc layer will change in the future.

You can use std::time::SystemTime (prefer way) or Duration since UNIX_EPOCH

ydb/src/grpc_wrapper/raw_topic_service/common/metering_mode.rs
```diff
@@ -0,0 +1,17 @@
+#[derive(serde::Serialize)]
+pub(crate) enum RawMeteringMode {
+    UseDefault,
```
rekby, line 3: It wil be better to use protospec defined names (with truncated prefixes), because it will be same words for same values in all places.

In this case it better to use Unspecified instead of UseDefault.

rekby:
Hello, thanks to your pr, good work.

It need only few small changes.

ydb/src/grpc_wrapper/raw_topic_service/common/consumer.rs
AlexKrudu: IMO storing Duration is better. 
Just so we don't have to deal with conversion errors here.

ydb/src/grpc_wrapper/raw_topic_service/common/codecs.rs
```diff
@@ -0,0 +1,36 @@
...
+impl From<RawSupportedCodecs> for SupportedCodecs {
+    fn from(value: RawSupportedCodecs) -> Self {
+        Self {
+            codecs: value
+                .codecs
+                .into_iter()
+                .map(|x| match x {
+                    RawCodec::Custom(val) => val,
+                    RawCodec::Unspecified => 0,
+                    RawCodec::Raw => 1,
```
rekby, line 28: Here need to use GRPC constants, becouse it is difficult to match integer constants by mind and very easy add bug by typo.

rekby (approved):

PR #133, opened 2023-03-05
feat: impl basic topic client, add create delete integration test

AlexKrudu:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

### DIFF
--- a/ydb/Cargo.toml
+++ b/ydb/Cargo.toml
@@ -11,6 +11,7 @@ repository = "https://github.com/ydb-platform/ydb-rs-sdk/tree/master/ydb"
 [dependencies]
 async-trait="0.1"
 derivative="2"
+derive_builder="0.12.0"
 http = "0.2"
 itertools="0.10"
 num = "0.4"
--- a/ydb/Cargo.toml
+++ b/ydb/Cargo.toml
@@ -1,12 +1,13 @@
 [package]
 publish=true
 name = "ydb"
-version = "0.5.4"
+version = "0.5.6"
 authors = ["rekby <timofey.koolin@gmail.com>"]
 edition = "2021"
 license = "Apache-2.0"
 description = "Crate contains generated low-level grpc code from YDB API protobuf, used as base for ydb crate"
 repository = "https://github.com/ydb-platform/ydb-rs-sdk/tree/master/ydb"
+rust-version = "1.56"
 
 [dependencies]
 async-trait="0.1"
--- a/ydb/src/client_topic/client.rs
+++ b/ydb/src/client_topic/client.rs
@@ -13,13 +13,13 @@ use std::collections::HashMap;
 #[builder(build_fn(error = "errors::YdbError"))]
 pub struct TopicOptions {
     // Use TopicOptionsBuilder
-    #[builder(default)]
-    pub metering_mode: MeteringMode,
+    #[builder(setter(strip_option), default)]
+    pub metering_mode: Option<MeteringMode>,
     #[builder(default)]
     pub min_active_partitions: i64,
     #[builder(default)]
     pub partition_count_limit: i64,
-    #[builder(default)]
+    #[builder(setter(strip_option), default)]
     pub retention_period: Option<core::time::Duration>,
     #[builder(default)]
     pub retention_storage_mb: i64,
@@ -62,7 +62,7 @@ impl TopicClient {
         path: String,
         topic_options: TopicOptions,
     ) -> YdbResult<()> {
-        let req = RawCreateTopicRequest::new(path, self.timeouts.operation_params(), topic_options)?;
+        let req = RawCreateTopicRequest::new(path, self.timeouts.operation_params(), topic_options);
 
         let mut service = self.connection().await?;
         service.create_topic(req).await?;
--- a/ydb/src/client_topic/client.rs
+++ b/ydb/src/client_topic/client.rs
@@ -1,0 +1,92 @@
+use crate::client::TimeoutSettings;
+use crate::client_topic::list_types::{Consumer, MeteringMode, SupportedCodecs};
+use crate::errors;
+use crate::grpc_connection_manager::GrpcConnectionManager;
+use crate::grpc_wrapper::raw_topic_service::create_topic::RawCreateTopicRequest;
+use crate::grpc_wrapper::raw_topic_service::delete_topic::RawDropTopicRequest;
+use crate::YdbError::InternalError;
+use crate::{grpc_wrapper, YdbResult};
+use derive_builder::{Builder, UninitializedFieldError};
+use std::collections::HashMap;
+
+#[derive(Builder)]
+#[builder(build_fn(error = "errors::YdbError"))]
+pub struct TopicOptions {
+    // Use TopicOptionsBuilder
+    #[builder(default)]
+    pub metering_mode: MeteringMode,
+    #[builder(default)]
+    pub min_active_partitions: i64,
+    #[builder(default)]
+    pub partition_count_limit: i64,
+    #[builder(default)]
+    pub retention_period: Option<core::time::Duration>,
+    #[builder(default)]
+    pub retention_storage_mb: i64,
+    #[builder(default)]
+    pub supported_codecs: SupportedCodecs,
+    #[builder(default)]
+    pub partition_write_speed_bytes_per_second: i64,
+    #[builder(default)]
+    pub partition_write_burst_bytes: i64,
+    #[builder(default)]
+    pub attributes: HashMap<String, String>,
+    #[builder(default)]
+    pub consumers: Vec<Consumer>,
+}
+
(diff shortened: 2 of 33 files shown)


ydb/src/grpc_wrapper/raw_topic_service/client.rs
```diff
@@ -17,11 +18,23 @@ impl RawTopicClient {
         }
     }
 
-    pub async fn create_topic(&mut self, req: RawCreateTopicRequest) -> RawResult<RawCreateTopicResult> {
-        request_with_result!(
+    pub async fn create_topic(&mut self, req: RawCreateTopicRequest) -> RawResult<()> {
+        request_without_result!(
```
AlexKrudu, line 22: if i try to use RawCreateTopicResult, https://github.com/ydb-platform/ydb-rs-sdk/blob/763f27cfdeaef432b6f822b202081bc4727196f7/ydb/src/grpc_wrapper/grpc.rs#L22 error gets thrown. I am not quite sure if I am do smth about it. My best guess is that happens cuz of result struct being empty

ydb/src/grpc_wrapper/raw_topic_service/common/codecs.rs
```diff
@@ -1,8 +1,8 @@
 use itertools::Itertools;
 use ydb_grpc::ydb_proto::topic::{Codec, SupportedCodecs};
 
-#[derive(serde::Serialize)]
-pub(crate) enum RawCodec {
+#[derive(serde::Serialize, Clone)]
+pub enum RawCodec {
```
AlexKrudu, line 5: since I am using this struct in a CreateTopicOptions, pub(create) mod had to be changed. I am not quite sure if this a good thing that Raw struct becomes pub. I just didn't want to copy-paste another struct just to put in user-space level i guess...

rekby: I think in the case copy-paste is good:

1. When we strong separate public interface from internals - it will easy to make change internals in the future. 
2. It prevent public some private API (for example internal errors from methods, From trait implementations, may be some methods).


And for public interface will be better have no strung difference between well-known and custom codecs: for better add codecs in the future.
And in code work with custom and well-known code will same (may be optional support default-implementation for well-known codec, but may be not for all and customer can re-define it).

```rust
struct Codec {
    pub code: i32
}

impl Codec {
    pub const RAW: Codec = Codec{code:1};
    pub const GZIP: Codec = Codec{code:2};

    pub fn is_custom(Codec: Codec)->bool{
        Codec.code >= 10000 && Codec.code < 20000
    }
}
```

ydb/src/client_topic/create_topic_options.rs
```diff
@@ -0,0 +1,37 @@
+use crate::grpc_wrapper::raw_topic_service::common::codecs::RawSupportedCodecs;
+use crate::grpc_wrapper::raw_topic_service::common::consumer::RawConsumer;
+use crate::grpc_wrapper::raw_topic_service::common::metering_mode::RawMeteringMode;
+use derive_builder::Builder;
+use std::collections::HashMap;
+use crate::errors;
+use crate::YdbError::{InternalError};
+
+#[derive(Builder)]
+pub struct TopicOptions { // Use TopicOptionsBuilder
```
rekby, line 10: It will be better move options to client.rs file - near with method.

It should have reason for separate it.
In go options separated for workarond of type/method visibility restrictions.
In rust crate most options placed near the method.

ydb/src/grpc_wrapper/raw_topic_service/common/consumer.rs
```diff
@@ -5,8 +5,8 @@ use crate::grpc_wrapper::raw_topic_service::common::codecs::RawSupportedCodecs;
 use ydb_grpc::google_proto_workaround::protobuf::Timestamp;
 use ydb_grpc::ydb_proto::topic::{Consumer, SupportedCodecs};
 
-#[derive(serde::Serialize)]
-pub(crate) struct RawConsumer {
+#[derive(serde::Serialize, Clone)]
+pub struct RawConsumer {
```
rekby, line 9: Raw types should be crate-only visibility for prevent difficults in the future.

ydb/src/grpc_wrapper/raw_topic_service/common/metering_mode.rs
```diff
@@ -1,5 +1,5 @@
-#[derive(serde::Serialize)]
-pub(crate) enum RawMeteringMode {
+#[derive(serde::Serialize, Clone)]
+pub enum RawMeteringMode {
```
rekby, line 2: some about visibility

ydb/src/lib.rs
```diff
@@ -82,6 +82,9 @@ pub(crate) mod transaction;
 mod types;
 mod types_converters;
 pub(crate) mod waiter;
+mod client_topic;
```
rekby, line 85: Add new public types to as "pub use ..." below please: By design of the crate all public-visibility types should be on top level and pub explicitly.

ydb/src/topic_create_delete_test.rs
```diff
@@ -0,0 +1,28 @@
...
+async fn create_delete_topic_test() -> YdbResult<()>{
+    let client = create_client().await?;
+    let database_path = client.database();
+    let topic_name = "test_topic".to_string();
+    let topic_path = format!("{}/{}", database_path, topic_name);
+
+    let mut topic_client = client.topic_client();
+    let mut scheme_client = client.scheme_client();
+
+    topic_client.create_topic(topic_path.clone(), TopicOptionsBuilder::default().build()?).await?;
```
rekby, line 19: it can have problem for run in not-empty (run on standalone database or persistent runned docker-container after previous run failed) and database will have "test_topic".

It will be better drop the topic before test (with ignore scheme error).

ydb/Cargo.toml
```diff
@@ -12,6 +12,7 @@ rust-version = "1.56"
 [dependencies]
 async-trait="0.1"
 derivative="2"
+derive_builder="0.12.0"
```
rekby, line 15: great

ydb/src/client_topic/list_types.rs
```diff
@@ -0,0 +1,107 @@
...
+        self.code >= 10000 && self.code < 20000
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct SupportedCodecs {
+    pub codecs: Vec<Codec>,
+}
+
+impl TryFrom<Codec> for RawCodec {
```
rekby, line 29: It no need error for conversion - user can set any codec. For example if YDB will publish "reverse" codec as well-known const = 5 and user have implementation for this codec (same way as for any custom codec). No reason for deny to use it.

It can implement simple From and direct copy codec value to Raw format.

Raw can detect well-known constants if someself (if need).

ydb/src/client_topic/client.rs
```diff
@@ -0,0 +1,92 @@
...
+use crate::{grpc_wrapper, YdbResult};
+use derive_builder::{Builder, UninitializedFieldError};
+use std::collections::HashMap;
+
+#[derive(Builder)]
+#[builder(build_fn(error = "errors::YdbError"))]
+pub struct TopicOptions {
+    // Use TopicOptionsBuilder
+    #[builder(default)]
+    pub metering_mode: MeteringMode,
```
rekby, line 17: What about use Option<MeteringMode> and None as unspecified? Will it more rust-native way?

ydb/src/grpc_wrapper/raw_topic_service/common/codecs.rs
```diff
@@ -11,7 +11,7 @@ pub(crate) enum RawCodec {
     Custom(i32),
 }
 
-#[derive(serde::Serialize)]
+#[derive(serde::Serialize, Clone, Default)]
 pub(crate) struct RawSupportedCodecs {
     pub codecs: Vec<RawCodec>,
 }
```
rekby, line 17: what about define fn RawCodec::from_int(codec: i32) and move map to the method?

```
                    RawCodec::Custom(val) => val,
                    RawCodec::Unspecified => Codec::Unspecified.into(),
                    RawCodec::Raw => Codec::Raw.into(),
                    RawCodec::Gzip => Codec::Gzip.into(),
                    RawCodec::Lzop => Codec::Lzop.into(),
                    RawCodec::Zstd => Codec::Zstd.into(),
```

it will allow deduplicate the map when convert from public Codec to RawCodec.

ydb/src/grpc_wrapper/raw_topic_service/client.rs
rekby: Server does not sent result for the operation, it can be bug or optimization.

Use without result now is ok.

rekby (approved):

PR #173, opened 2024-02-12
password authentication

uslon:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


ydb/src/auth_test.rs
```diff
@@ -0,0 +1,9 @@
+use tracing_test::traced_test;
+
+use crate::YdbResult;
+
+#[tokio::test]
+#[traced_test]
+async fn auth_test() -> YdbResult<()> {
+    Ok(())
```
rekby, line 8: Need real tests:
1. "SELECT 1" with good static_credentials - must be ok
2. "SELECT 1" with bad static_credentials (use bad password for user) - must failed

ydb/src/client_builder.rs
```diff
@@ -25,6 +25,7 @@ static PARAM_HANDLERS: Lazy<Mutex<HashMap<String, ParamHandler>>> = Lazy::new(||
         m.insert("token".to_string(), token);
         m.insert("token_cmd".to_string(), token_cmd);
         m.insert("token_metadata".to_string(), token_metadata);
+        m.insert("token_password".to_string(), token_password);
```
rekby, line 28: Better use `password` for query arg

ydb/src/grpc_wrapper/raw_auth_service/client.rs
```diff
@@ -4,23 +4,23 @@ use crate::grpc_wrapper::raw_auth_service::login::RawLoginResult;
...
 impl RawAuthClient {
-    #[allow(dead_code)]
     pub fn new(service: InterceptedChannel) -> Self {
         Self {
             service: AuthServiceClient::new(service),
         }
     }
 
-    #[allow(dead_code)]
+    #[tokio::main(flavor = "current_thread")]
```
rekby, line 23: it is wrapper only, the code must be work with any tokion flavour.

Use `#[tokio::main(flavor = "current_thread")]` is good idea, but not very good place.

You should create runtime (use the attribute) on border between async and sync code. In that case - better create additional function in UserPasswordAuth struct as proxy (for create async runtime).

ydb/Cargo.toml
```diff
@@ -16,6 +16,7 @@ force-exhaustive-all=[] # The feature disable all non_exhaustive attributes in y
 async-trait="0.1"
 derivative="2"
 derive_builder="0.12.0"
+futures = "0.3"
```
rekby, line 19: The sdk has dependency from tokio already. Use second async runtime may cause additional hard debug problems.

It is a good approach as first step - for implement "work concept".
Second step - remove the additional dependency and use already present dependency - tokio.

ydb/src/auth_test.rs
```diff
@@ -0,0 +1,88 @@
...
+#[traced_test]
+#[ignore] // YDB access is necessary
+async fn password_client_test() -> YdbResult<()> {
+    let client = create_password_client().await?;
+
+    let sum: i32 = client
+    .table_client() // create table client
+    .retry_transaction(|mut t: Box<dyn Transaction>| async move {
+        // send the query to the database
+        let res = t.query(Query::from("SELECT 17 + 25 AS sum")).await?;
```
rekby, line 75: What about to replace query to `SELECT 1`? 
It is best case for check "minimal query works".

ydb/src/client_builder.rs
```diff
@@ -25,6 +25,7 @@ static PARAM_HANDLERS: Lazy<Mutex<HashMap<String, ParamHandler>>> = Lazy::new(||
         m.insert("token".to_string(), token);
         m.insert("token_cmd".to_string(), token_cmd);
         m.insert("token_metadata".to_string(), token_metadata);
+        m.insert("token_password".to_string(), password);
```
rekby, line 28: better to replace it for "token_static_password"
where static - is approach for receive token and password is param for static type

ydb/src/client_builder.rs
```diff
@@ -101,6 +102,48 @@ fn token_metadata(uri: &str, mut client_builder: ClientBuilder) -> YdbResult<Cli
     Ok(client_builder)
 }
 
+fn password(uri: &str, mut client_builder: ClientBuilder) -> YdbResult<ClientBuilder> {
+    let mut username = Option::<String>::default();
+    let mut password = Option::<String>::default();
+
+    for (key, value) in url::Url::parse(uri)?.query_pairs() {
+        match key.as_ref() {
+            "password" => {
```
rekby, line 111: A command for detect static credentials type auth is `token_password` (or `token_static_password` in future). 
Now user must set both fields `token_password` and `password`: first for enable the handler and second for set password.

Need consistent parameter name: token_static_login and token_static_password.

ydb/src/credentials.rs
```diff
@@ -228,3 +235,48 @@ impl Credentials for GCEMetadata {
         format!("GoogleComputeEngineMetadata from {}", self.uri.as_str())
     }
 }
+
+pub struct UserPasswordAuth {
```
rekby, line 239: Better to rename the structure to StaticCredentialsAuth - by auth type name (we named it static credentials)

rekby (changes requested):
It is good work, thanks.
It needs only small changes

ydb/src/client_builder.rs
uslon: Maybe it's better to use `token_static_username` as a parameter name?

rekby:
@uslon Thanks for the good works

PR #183, opened 2024-04-01
Service account credentials (key)

MikhailNazarov:

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [x] Other (please describe): add dependency to jsonwebtoken for jwt and chrono for parsing token expiration

There is no credential provider for service account key json file.
reqwest not allow https requests saying that only http is allowed

## What is the new behavior?

Implemented ServiceAccountCredentials
Also exampl auth-ycloud-serviceaccount was provided  for usage demonstation and debugging.

changed reqwest feature from rustls to rustls-tls fixed issue of sending https requests.

## Other information

 Example:
 ```
 use ydb::ServiceAccountCredentials;
 let cred = ServiceAccountCredentials::from_env()?;
 ```
 or
 ```
 let json = "....";
 let cred = ServiceAccountCredentials::from_json(json)?;
 ```
 or
 ```
 let cred = ServiceAccountCredentials::new("service_account_id", "key_id", "private_key");
 ```
 or
 ```
 let cred = ServiceAccountCredentials::from_file("/path/to/file")?;
 ```


ydb/src/credentials.rs
```diff
@@ -147,6 +151,163 @@ impl Credentials for CommandLineYcToken {
...
+        let key: JsonKey = serde_json::from_str(json_key)?;
+
+        Ok(Self {
+            key_id: key.id,
+            service_account_id: key.service_account_id,
+            private_key: SecretString::new(key.private_key),
+        })
+    }
+
+    const AUDIENCE: &'static str = "https://iam.api.cloud.yandex.net/iam/v1/tokens";
```
rekby, line 229: Please make it as config param for `ServiceAccountCredentials` with the default value.
IAM service may have more, then one installation/endpoint, similar to [google/yandex metadata request uri](https://github.com/ydb-platform/ydb-rs-sdk/blob/main/ydb/src/credentials.rs#L333-L333).

ydb/src/credentials.rs
```diff
@@ -147,6 +151,163 @@ impl Credentials for CommandLineYcToken {
...
+        let req = TokenRequest { jwt };
+        let client = reqwest::blocking::Client::new();
+        let res: TokenResponse = client
+            .request(reqwest::Method::POST, API_URL)
+            .json(&req)
+            .send()?
+            .json()?;
+
+        Ok(TokenInfo::token(format!("Bearer {}", res.iam_token))
+            .with_renew(Self::time_to_instant(res.expires_at)))
```
rekby, line 307: IAM has recommend for [renew token before it expires](https://yandex.cloud/en/docs/iam/concepts/authorization/iam-token).

It allow to prevent problems:
1. Time skew between IAM and client system.
2. Additional latency for wait new certificate
3. Continue to work if iam service failed on renew certificate, while IAM team repairing the service.

By common rule: we request IAM certificate for 11-12 hours and renew the token while it has about 90% of lifetime (in that case - after one hour of token usage).

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
rekby, line 1: Remove `.vscode` folder from the PR please.

Cargo.lock
```diff
@@ -44,7 +44,7 @@ checksum = "648ed8c8d2ce5409ccd57453d9d1b214b342a0d69376a6feda1fd6cae3299308"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.109",
```
rekby, line 47: Do the changes really require a lot upgrade of dependencies or it happened accidentally?

If it - it is OK.
If it accidentally - will be better to revert unneed updates - for continue to test with as older libraries as possible (because upgrade at client code much easer then downgrade if it needs).

rekby (changes requested):
Hello @MikhailNazarov.

Thanks for great work!

I have few comments, for do the PR even better.

ydb/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-publish=true
+publish = true
```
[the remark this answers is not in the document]
MikhailNazarov: I use VSCode, rust code formatted with rust-analyzer, toml with Even Better Toml (I think so). I'll try to investigate how to do format in cli

ydb/Cargo.toml
```diff
@@ -10,40 +10,51 @@ repository = "https://github.com/ydb-platform/ydb-rs-sdk/tree/master/ydb"
...
+jsonwebtoken = "7.2.0"
 num = "0.4"
 mockall = "0.10"
 once_cell = "1.8"
 prost = "0.11.2"
 prost-types = "0.11.2"
 pbjson-types = "0.5.1"
 rand = "0.8"
-reqwest = {version="0.11", features = ["blocking", "json","rustls"], default-features = false}
+reqwest = { version = "=0.11.9", features = [
```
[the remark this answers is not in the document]
MikhailNazarov: [x] reverted to version 0.11. everything still works.

ydb/examples/auth-ycloud-serviceaccount.rs
```diff
@@ -0,0 +1,40 @@
...
+#[tokio::main]
+async fn main() -> YdbResult<()> {
+    init_logs();
+    info!("Building client");
+
+    let connection_string =
+        env::var("YDB_CONNECTION_STRING").expect("YDB_CONNECTION_STRING not set");
+
+    let client = ClientBuilder::new_from_connection_string(connection_string)?
+        .with_credentials(ServiceAccountCredentials::from_env().unwrap())
```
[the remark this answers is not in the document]
MikhailNazarov: - [x] fixed

ydb-grpc/src/generated/google.protobuf.rs
```diff
@@ -17,19 +17,15 @@
 ///      if (any.is(Foo.class)) {
 ///        foo = any.unpack(Foo.class);
 ///      }
-///      // or ...
-///      if (any.isSameTypeAs(Foo.getDefaultInstance())) {
-///        foo = any.unpack(Foo.getDefaultInstance());
-///      }
-///   Example 3: Pack and unpack a message in Python.
+/// Example 3: Pack and unpack a message in Python.
```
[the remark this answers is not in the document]
MikhailNazarov: runed specified command. Now it's ok?

ydb/src/credentials.rs
MikhailNazarov: I maked this value as parameter with default value. But little bit different from metadata, because there is multiple methods for creating instance. I added separate method to specify custom url:

```
 pub fn with_url(mut self, url: impl Into<String>) -> Self {
     self.audience_url = url.into();
     self
}
```

.vscode/settings.json
MikhailNazarov: - [x] done

Cargo.lock
MikhailNazarov: this change is happening automatic after build. Maybe on of the dependencies required it

ydb/src/credentials.rs
MikhailNazarov: Ok. Am I required to replace the original token lifetime value with 720 instead of 3600? in create_token(), I return by analogy with GCEMetadata. Then, somewhere in the code, a decision is made about 90%?

ydb/Cargo.toml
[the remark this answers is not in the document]
MikhailNazarov: taplo-cli can format toml-files very good. Better toml uses it.
It can be installed or via package manager as binary package or via cargo install taplo-cli

to format Cargo.toml we can use, for example, 

```
 taplo fmt --option indent_string="    "  Cargo.toml 
```

options can be located in separate config file or placed into the command string.

rekby (approved):

PR #196, opened 2024-05-06
Support custom CA certificates for TLS

uslon:

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-


ydb/src/credentials.rs
```diff
@@ -557,12 +559,15 @@ impl StaticCredentials {
         Ok(raw_response.token)
     }
 
-    pub fn new(username: String, password: String, endpoint: Uri, database: String) -> Self {
+    pub fn new(username: String,
```
rekby, line 562: it will be better create new method: `new_with_ca` instead of modify method `new`:
- no break api
- allow use cert param only if need. It isn't need if connection doesn't need TLS or TLS has certificate from os-trusted ca.
-

ydb/src/custom_ca_test.rs
```diff
@@ -0,0 +1,25 @@
...
+    // Query,
+    // Transaction,
+    YdbResult,
+};
+
+#[tokio::test]
+#[traced_test]
+#[ignore] // YDB access is necessary
+async fn custom_ca_test() -> YdbResult<()> {
+    // let client = create_custom_ca_client().await?;
```
rekby, line 13: it will be better to "return ()" at start with comment - for check compilation even if we can't check work now.

ydb/src/connection_pool.rs
```diff
@@ -1,20 +1,37 @@
...
 impl ConnectionPool {
     pub(crate) fn new() -> Self {
         Self {
             state: Arc::new(Mutex::new(ConnectionPoolState::new())),
+            tls_config: None.into(),
+        }
+    }
+
+    pub(crate) fn load_certificate(self, path: String) -> Self {
+        trace!("path is {}", path);
```
rekby, line 25: better merge it to one record. At real world system can generate a lot of logs even at start (with multiply treads) and two records may be separated by many other records if trace write as logs.

rekby (changes requested):

rekby (approved):

PR #206, opened 2024-08-04
Nearest dc balancer

bakalover:


bakalover:
- [ ] сквош

ydb/src/load_balancer.rs
```diff
@@ -1,16 +1,27 @@
...
+use std::collections::HashMap;
+use std::net::ToSocketAddrs;
 use std::sync::{Arc, RwLock};
 use tokio::sync::watch::Receiver;
 
 #[mockall::automock]
 pub(crate) trait LoadBalancer: Send + Sync + Waiter {
     fn endpoint(&self, service: Service) -> YdbResult<Uri>;
     fn set_discovery_state(&mut self, discovery_state: &Arc<DiscoveryState>) -> YdbResult<()>;
+    fn set_config(&mut self, config: BalancerConfig) -> YdbResult<()>;
```
rekby, line 24: The config of balancer is static info, it doesn't need changed in runtime and no need the method in the create.
Try to create balancer with config at start.

It allow to implement different config for different balancer, no need to think about possibility to change config on the fly, etc, no need return errors from static balancer and handle the error.

ydb/src/load_balancer.rs
```diff
@@ -182,6 +215,247 @@ pub(crate) async fn update_load_balancer(
...
+// What will balancer do if there is no available endpoints at local dc
+pub(crate) enum FallbackStrategy {
+    Error,  // Just throw error
+    Random, // Random endpoint from other dcs
+}
+
+impl Default for BalancerConfig {
+    fn default() -> Self {
+        BalancerConfig {
+            fallback_strategy: FallbackStrategy::Error,
```
rekby, line 231: why the error is default fallback?

ydb/src/load_balancer.rs
```diff
@@ -182,6 +215,247 @@ pub(crate) async fn update_load_balancer(
     }
 }
 
+pub(crate) struct BalancerConfig {
+    fallback_strategy: FallbackStrategy,
+}
+
+// What will balancer do if there is no available endpoints at local dc
+pub(crate) enum FallbackStrategy {
```
rekby, line 223: What about use other balancer as fallback?

Or use two balancers: for balancing nodes in nearest DC and fallback balancer.

ydb/src/load_balancer.rs
```diff
@@ -182,6 +215,247 @@ pub(crate) async fn update_load_balancer(
...
+        dc_to_nodes
+            .iter_mut()
+            .for_each(|(_, endpoints)| to_check.append(self.get_random_endpoints(endpoints)));
+        let local_dc = self.find_local_dc(&to_check)?;
+        self.location = local_dc;
+        Ok(())
+    }
+
+    fn adjust_allowed_endpoints(&mut self) -> YdbResult<()> {
+        self.allowed_endpoints = self
```
rekby, line 313: what about change "allowed" to "prefer"?

ydb/src/load_balancer.rs
```diff
@@ -182,6 +215,247 @@ pub(crate) async fn update_load_balancer(
...
+                        }
+                    }
+                    _ = stop_measure.cancelled() => {
+                        ();
+                    }
+                }
+            });
+        }
+
+        tokio::task::block_in_place(move || {
```
rekby, line 426: Is it temporary block - for draft only?

rekby (commented):
Hello, thanks for the draft :)

I suggest few ideas for that state

ydb/src/load_balancer.rs
bakalover: Is there any better option to keep` LoadBalancer` trait's methods synchronous? Maybe blocking channel recieve?

bakalover: Otherwise, is it ok, if I will not wait to balancer to set it's location on `set_discovery_state`, because it may cause stale information on `endpoint` call

ydb/src/load_balancer.rs
```diff
@@ -312,4 +706,473 @@ mod test {
         assert!(*map.get(two.to_string().as_str()).unwrap() > 30);
         Ok(())
     }
+
+    #[test]
+    fn split_by_location() -> YdbResult<()> {
```
rekby, line 711: what about extract test to separate file?

it will be better to separate test/production code for search usages and diffs in PRs/

ydb/src/load_balancer.rs
```diff
@@ -312,4 +706,473 @@ mod test {
...
+        let nodes = vec![
+            NodeInfo::new(Uri::from_str("http://one:213")?, "A".to_string()),
+            NodeInfo::new(Uri::from_str("http://two:213")?, "A".to_string()),
+            NodeInfo::new(Uri::from_str("http://three:213")?, "B".to_string()),
+            NodeInfo::new(Uri::from_str("http://four:213")?, "B".to_string()),
+            NodeInfo::new(Uri::from_str("http://five:213")?, "C".to_string()),
+        ];
+        let splitted = NearestDCBalancer::split_endpoints_by_location(&nodes);
+        assert_eq!(splitted.keys().len(), 3);
```
rekby, line 720: What about compare with exact map instead of check lens?

What about the code push good numbers of items, but all items is bad uris, or same uri?

ydb/src/load_balancer.rs
```diff
@@ -312,4 +706,473 @@ mod test {
...
+            NodeInfo::new(Uri::from_str("http://nine:213")?, "C".to_string()),
+        ];
+
+        let mut refs = nodes.iter().collect_vec();
+        let nodes_clone = refs.clone();
+        let random_subset = NearestDCBalancer::get_random_endpoints(&mut refs);
+
+        assert_eq!(random_subset.len(), NODES_PER_DC);
+        for node in random_subset {
+            assert_true!(nodes_clone.contains(node))
```
rekby, line 746: check about all nodes are different

ydb/src/load_balancer.rs
```diff
@@ -182,17 +199,394 @@ pub(crate) async fn update_load_balancer(
...
+
+#[derive(Default)]
+struct BalancerState {
+    preferred_endpoints: Vec<NodeInfo>,
+}
+
+// What will balancer do if there is no available endpoints at local dc
+#[derive(PartialEq, Eq)]
+pub(crate) enum FallbackStrategy {
+    Error,            // Just throw error
```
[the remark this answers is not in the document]
bakalover: > what about use optional instead of enum?

Enum is extensible

ydb/src/load_balancer.rs
```diff
@@ -182,17 +199,394 @@ pub(crate) async fn update_load_balancer(
...
+        let self_waiter = Box::new(self.waiter.clone());
+        match self.config.fallback_balancer.as_ref() {
+            Some(balancer) => Box::new(AllWaiter::new(vec![self_waiter, balancer.waiter()])),
+            None => self_waiter,
+        }
+    }
+}
+
+const NODES_PER_DC: usize = 5;
+const PING_TIMEOUT_SECS: u64 = 60;
```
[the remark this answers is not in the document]
bakalover: 60 sec - interval for TimerDiscovery renewal (there is also no predefined constants)
https://github.com/ydb-platform/ydb-rs-sdk/blob/master/ydb/src/client_builder.rs#L283

ydb/src/load_balancer.rs
```diff
@@ -182,17 +199,394 @@ pub(crate) async fn update_load_balancer(
...
+
+impl NearestDCBalancer {
+    fn get_endpoint(&self, service: Service) -> YdbResult<Uri> {
+        match self.balancer_state.try_lock() {
+            Ok(state_guard) => {
+                for ep in state_guard.borrow().preferred_endpoints.iter() {
```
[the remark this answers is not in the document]
bakalover: Now it chooses randomly

ydb/src/load_balancer/nearest_dc_balancer.rs
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+}
+
+pub(super) const NODES_PER_DC: usize = 5;
+pub(super) const PING_TIMEOUT_SECS: u64 = 60;
+
+impl NearestDCBalancer {
+    fn get_endpoint(&self, service: Service) -> YdbResult<Uri> {
+        match self.balancer_state.try_lock() {
```
rekby, line 147: use lock instead of try_lock: we don`t need fast fail at the path: if code doesn't see endpoint it will return to retrier. The retraier will see custom error and fail transaction.

Better wait until update state finish: it is fast operation.

the method should return a error is the ballancer is empty

ydb/src/load_balancer/nearest_dc_balancer.rs
```diff
@@ -0,0 +1,388 @@
...
+            match Self::extract_nodes(&new_discovery_state) {
+                Ok(some_nodes) => {
+                    let mut dc_to_nodes = Self::split_endpoints_by_location(some_nodes);
+                    let mut to_check = Vec::with_capacity(NODES_PER_DC * dc_to_nodes.keys().len());
+                    dc_to_nodes.values_mut().for_each(|endpoints| {
+                        to_check.append(Self::get_random_endpoints(endpoints))
+                    });
+                    match Self::find_local_dc(&to_check).await {
+                        Ok(dc) => {
+                            info!("found new local dc:{}", dc);
```
rekby, line 201: it is often message, not intresting when all ok. Better to use debug for the message

ydb/src/load_balancer/nearest_dc_balancer.rs
```diff
@@ -0,0 +1,388 @@
...
+    }
+}
+
+pub(crate) struct NearestDCBalancer {
+    discovery_state: Arc<DiscoveryState>,
+    state_sender: Sender<Arc<DiscoveryState>>,
+    ping_token: CancellationToken,
+    waiter: Arc<WaiterImpl>,
+    config: BalancerConfig,
+    balancer_state: Arc<Mutex<BalancerState>>,
```
rekby, line 64: what about use RWLock instead of Mutex?
it will allow multiply reads in parallel and block for update state only

ydb/src/load_balancer/nearest_dc_balancer.rs
```diff
@@ -0,0 +1,388 @@
...
+            }
+        }
+    }
+
+    async fn adjust_preferred_endpoints(
+        balancer_state: &Arc<Mutex<BalancerState>>,
+        new_nodes: &Vec<NodeInfo>,
+        local_dc: String,
+    ) {
+        info!("adjusting endpoints");
```
rekby, line 221: it should be debug or trace log level

ydb/src/load_balancer/nearest_dc_balancer.rs
```diff
@@ -0,0 +1,388 @@
...
+                }
+            });
+        }
+
+        let wait_first_some_or_cancel = async {
+            loop {
+                tokio::select! {
+                    biased; // check timeout first
+                    _ = interrupt_collector_future.cancelled() =>{
+                        Self::join_all(&mut nursery).await; // Children will be cancelled due to tokens chaining, see (*)
```
rekby, line 357: what mean the comment: "see (*)"?

rekby (commented):
thanks for the great fork

I have only few small comments

ydb/src/load_balancer/nearest_dc_balancer.rs
bakalover: Because there is only reader and only one writer

ydb/src/load_balancer/nearest_dc_balancer.rs
bakalover: Changed to `blocking_lock()`

ydb/src/load_balancer/nearest_dc_balancer.rs
```diff
@@ -0,0 +1,390 @@
...
+        addrs: Vec<&String>,
+        timeout: Duration,
+    ) -> YdbResult<String> {
+        // Cancellation flow: timeout -> address collector -> address producers
+        let interrupt_via_timeout = CancellationToken::new();
+        let interrupt_collector_future = interrupt_via_timeout.child_token();
+        let stop_measure = interrupt_collector_future.child_token();
+
+        let (start_measure, _) = broadcast::channel::<()>(1);
+        let buffer_cap = if !addrs.is_empty() { addrs.len() } else { 1 };
```
rekby, line 327: better reverse for positive condition (for easer read):

```rust
if addrs.is_empty() { 1 } else {addrs.len()};
```

ydb/src/load_balancer/nearest_dc_balancer.rs
```diff
@@ -0,0 +1,390 @@
...
+        BalancerConfig {
+            fallback_strategy: FallbackStrategy::BalanceWithOther(Box::new(
+                RandomLoadBalancer::new(),
+            )),
+        }
+    }
+}
+
+pub(crate) struct NearestDCBalancer {
+    discovery_state: Arc<DiscoveryState>,
```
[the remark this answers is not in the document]
bakalover: removed

rekby (approved):
thanks for great work :)

PR #215, opened 2024-11-11
Add query statistics support

MikhailNazarov:
## Pull request type

- [x] Feature

## Other information

there are two tests:
- test_stats_updates
- test_stats_reads

first failed because no table access stats, second failed because it show two phases: update and read (?) when i send only second query with stats collection

### DIFF
--- a/ydb/src/client_table_test_integration.rs
+++ b/ydb/src/client_table_test_integration.rs
@@ -877,3 +877,91 @@ FROM
     assert!(result_set_count > 1); // ensure get multiply results
     Ok(())
 }
+#[tokio::test]
+#[traced_test]
+#[ignore] // need YDB access
+async fn test_stats_updates() -> YdbResult<()> {
+    let client = create_client()
+        .await?;
+
+    client
+        .table_client()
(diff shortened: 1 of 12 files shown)


ydb/src/client_table_test_integration.rs
```diff
@@ -877,3 +877,91 @@ FROM
...
+    
+    let res = tx.query(Query::new("SELECT * FROM test_values")
+    .with_stats(crate::QueryStatsMode::Basic))
+        .await?;
+
+    tx.commit().await?;
+
+    println!("{:?}", res.stats);
+
+    assert_eq!(2, res.stats.map(|x|x.rows_affected()).unwrap_or(0));
```
rekby, line 955: The reason same as above: upsert delay apply until ydb need it. In the case - up to the select. And upsert really apply only on when you send select.

You can select twice: upsert will be applied for first select and you will see only reads for second select.

rekby:
PR устарел :( 

SDK перешёл на query-сервис, а возможность выполнения запросов через table-сервис убрана совсем

PR #224, opened 2025-03-29
`cargo fmt` applied + `cargo fmt --check` in GitHub Actions

albenik:
Rust has an excellent built-in code formatter with good default settings.
When code in a project is properly formatted by default, each contributor can easily format their code according to Rust standards without the risk of affecting other parts of the project.
Otherwise, inconsistencies in formatting can lead to noisy diffs, unnecessary merge conflicts, and reduced overall code quality.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

`cargo fmt --check` returns error

Issue Number: N/A

## What is the new behavior?

`cargo fmt --check` returns no errors

## Other information

Just applied `cargo fmt` to code base, plus added `cargo fmt --check` step to Test workflow


.github/workflows/rust-tests.yml
```diff
@@ -59,6 +59,9 @@ jobs:
     - name: Rust cache
       uses: Swatinem/rust-cache@v2
 
+    - name: Rustfmt check
```
rekby, line 62: What about add condition for run the check for new rust only - for prevent conflict beween rust fmt rules if some rules will be change in the future?

rekby (commented):
Thanks for the PR, it is very good to check code style. 
I have a suggestion for the check

.github/workflows/rust-tests.yml
albenik: This looks like premature optimization.

Let’s consider scenarios where formatting rules could change and break the current check:

1. The Rust team changes the default `rustfmt` rules in a new version (if I understood correctly, this is what you're referring to).
2. The maintainers decide to change the code style and add a `.rustfmt.toml` file with custom rules.
3. Someone submits a PR that accidentally or intentionally includes a `.rustfmt.toml` with custom settings.

The first case seems unlikely. But even if it happens, it’s better to immediately see an error and then add a `.rustfmt.toml` to preserve the previous formatting rules — rather than end up with silent code reformatting and backward-incompatible `cargo fmt` behavior across the project.

In the second case, if the maintainers intentionally change the style, they’ll likely apply `cargo fmt` across the whole codebase. It would be strange if they didn’t.

In the third case, if we only check the modified code, we might not detect that the rules have changed — which could lead to unnoticed issues.

albenik: I was in such a hurry to reply that I didn't notice the request had already been merged :)

Issue #359, opened 2025-07-13
dev: topicwriter: Using immutable topic writer for write()


lobart:
**Impact Assessment**

How would this change affect:
- [ ] Performance
- [ ] Maintainability
- [ ] Backward compatibility
- [x] User experience

**Area of Improvement**

I want to improve TopicWriter for using write() method without requirement of TopicWriter mutability.
This question is highlighted in comment https://github.com/ydb-platform/ydb-rs-sdk/blob/fd92cfa274be104c7f079356406acedc9837c5d7/ydb/examples/topic-writer.rs#L22 

**Current Situation**

Now TopicWriter have to be mutable for using write() method.

**Suggested Improvement**

Change field last_seq_num_handled of TopicWriter from i64 to AtomicI64.
And for all calls lock() in writer.rs add except(POISONED_MUTEX) instead unwrap().

PR #360, opened 2025-07-13
Refactor topic writer: let using write() with immutable TopicWriter

lobart:
Change type of field last_seq_num_handled from i64 to AtomicI64

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Requirement of mutable TopicWriter for using method write().
Using unwrap() after calls of lock().

Issue Number: 359
https://github.com/ydb-platform/ydb-rs-sdk/issues/359

## What is the new behavior?

- calling write() method doesn't require mutability of TopicWriter 
- expect(POISON_MUTEX) is used instead lock().unwrap(), where POISON_MUTEX is error description for all errors related to poisoned mutexes in writer.rs.


rekby:
Hello.

Thanks for the PR.
Can you explain - why do you need write as unmutable method?

PR #480, opened 2026-06-18
slo: tune rate limiter

asmyasnikov:
## Summary
- Align Rust SLO rate limiter with Go: `rate.NewLimiter(rps, 1)` — burst=1, one token at start.
- Replaces `max_tokens(rps)` / `initial_available=0` with per-interval refill of 1 token.

## Hypothesis
Rate limiter semantics mismatch may explain ~900 vs ~1000 read ops/s once SDK latency is fixed. At current ~1.8s op latency, throughput should stay ~400–500 ops/s (limiter not binding).

## Test plan
- [ ] Add `SLO` label to run ydb-slo-action
- [ ] Compare read/write throughput and P50 vs master baseline
- [ ] Expect no improvement at current SDK latency; after SDK fix, should approach Go ~1000 read ops/s

Made with [Cursor](https://cursor.com)


tests/slo/slo-framework/src/helpers.rs
```diff
@@ -5,10 +5,15 @@ use ratelimit::Ratelimiter;
 use tokio::time::sleep;
 use tokio_util::sync::CancellationToken;
 
+/// Shared rate limiter aligned with Go SLO: `rate.NewLimiter(rate.Limit(rps), 1)`.
+///
+/// Refills 1 token every `1/rps` seconds, burst capacity 1, one token available at start.
```
asmyasnikov:
```suggestion
```

Issue #427, opened 2026-06-10
topic-service: Batching on Read


asmyasnikov:

sshaplygin:
Has the issue been resolved?

aral01:
Closed in #535

PR #504, opened 2026-06-29
TopicReader messages batching

aral01:
Batching messages in TopicReader by partition_session_id, fair chose via RoundRobin, support for EndPartititonRequest from server.

## Pull request type

This PR depends on #503 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

TopicReader inner buffer is VecDeque, which stores all messages sequentially. Batch is created by collecting messages with same partition session id from the start of the deque. 

This approach naturally supports EndPartition guarantees, but request was not implemented.

Issue Number: #427

## What is the new behavior?

- Batching is done via storing messages in vecdeques separately for each partition via hashmap.
- Fairness guarantees are achieved via RoundRobin algorithm on partition session ids
- EndPartition request is supported via registration mechanics: each parent registers him and his childs. Childs are not allowed to enter RoundRobin container, until parent messages are not exhausted.


aral01:
Moved to #510

PR #540, opened 2026-07-13
Fix: SLO e2e metrics and reliability

aral01:
## Pull request type

  - [x] Bugfix
  - [ ] Feature
  - [ ] Build related changes

  Topic SLO records message RTT as a generic operation latency, so it is not included
  in the resulting statistics.

  Some operation metrics are also reported too early: writes are counted before
  write acknowledgement, and reads before commit acknowledgement. Worker failures
  and invariant violations may be lost during shutdown.

  Issue Number: N/A

  ## What is the new behavior?

  - Added `topic_e2e` p50/p95/p99 metrics to the SLO report
  - Topic e2e latency is measured from message creation to processing by the reader
  - Write success is recorded after write acknowledgement
  - Read success is recorded after commit acknowledgement
  - Topic workers are supervised; worker and invariant errors fail the workload
  - Removed `process::exit`, so metrics and other resources are shut down normally
  - Replaced the bursty rate limiter with a fair async limiter
  - All percentiles use the same OTLP export snapshot and are reset after export
  - Long latency outliers are no longer silently dropped
  - Wired topic e2e metrics into CI

  ## Other information

  For a successful topic SLO run this PR also needs:

  - #539 — fixes the discovery deadlock which can hang the workload
  - #535 — fixes topic reader batching and SLO invariant breakage

### DIFF
--- a/tests/slo/slo-framework/Cargo.toml
+++ b/tests/slo/slo-framework/Cargo.toml
@@ -21,6 +21,9 @@ opentelemetry_sdk = { version = "0.27", features = ["rt-tokio"] }
 opentelemetry-semantic-conventions = "0.27"
 rand = "0.8.5"
 rand_core = "0.6.4"
-tokio = { version = "=1.38.1", features = ["full", "signal", "test-util"] }
+tokio = { version = "=1.38.1", features = ["full", "signal"] }
 tokio-util = { version = "=0.7.11" }
 ydb = { path = "../../../ydb" }
+
+[dev-dependencies]
+tokio = { version = "=1.38.1", features = ["test-util"] }
--- a/tests/slo/slo-framework/Cargo.toml
+++ b/tests/slo/slo-framework/Cargo.toml
@@ -21,7 +21,6 @@ opentelemetry_sdk = { version = "0.27", features = ["rt-tokio"] }
 opentelemetry-semantic-conventions = "0.27"
 rand = "0.8.5"
 rand_core = "0.6.4"
-ratelimit = "0.10.0"
-tokio = { version = "=1.38.1", features = ["full", "signal"] }
+tokio = { version = "=1.38.1", features = ["full", "signal", "test-util"] }
 tokio-util = { version = "=0.7.11" }
 ydb = { path = "../../../ydb" }
--- a/tests/slo/slo-framework/src/topic/workload.rs
+++ b/tests/slo/slo-framework/src/topic/workload.rs
@@ -170,6 +170,7 @@ async fn writer_worker(
                 let msg = err.to_string();
                 span.finish(Some(&msg), 1);
                 fw.logger.errorf(format!("write failed: {msg}"));
+                return Err(format!("writer failed: {msg}"));
             }
             Err(_) => {
                 span.finish(Some("write timeout"), 1);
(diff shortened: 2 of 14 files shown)


tests/slo/slo-framework/src/topic/workload.rs
```diff
@@ -80,124 +86,157 @@ impl<T: TopicService + 'static> Workload for TopicWorkload<T> {
...
+    loop {
+        let delivery_span = fw.metrics.start(OPERATION_READ);
+        let mut batch = match timeout(delivery_timeout, reader.read_batch()).await {
+            Ok(Ok(batch)) => batch,
+            Ok(Err(err)) => {
+                let msg = err.to_string();
+                delivery_span.finish(Some(&msg), 1);
+                fw.logger.errorf(format!("read failed: {msg}"));
+                return Err(format!("reader {worker_id} failed: {msg}"));
             }
```
rekby, line 201: A transient SDK read error here aborts the **entire** workload (the returned `Err` propagates through `supervise_workers` and fails the run). Combined with the new `fail_on_workload_error: "true"` in `slo.yml`, a single read blip during a ~600s run will fail CI.

This is asymmetric with `writer_worker`, where a `write_with_ack` error is only recorded via `span.finish(Some(&msg), 1)` and the loop continues — a write worker never fails the workload. A delivery timeout on the read side is also tolerated (`continue`). So only a non-timeout read error is treated as fatal.

Is this asymmetry intended? For an SLO reliability workload you usually want to *record* the failed read (which is already done on the line above) and keep measuring the success rate, reserving hard workload failure for invariant violations (`process_batch` errors), which are genuinely unrecoverable data-integrity problems. If the intent really is "any read error is fatal," it may be worth applying the same rule to writes for consistency.

tests/slo/slo-framework/Cargo.toml
```diff
@@ -21,7 +21,6 @@ opentelemetry_sdk = { version = "0.27", features = ["rt-tokio"] }
 opentelemetry-semantic-conventions = "0.27"
 rand = "0.8.5"
 rand_core = "0.6.4"
-ratelimit = "0.10.0"
-tokio = { version = "=1.38.1", features = ["full", "signal"] }
+tokio = { version = "=1.38.1", features = ["full", "signal", "test-util"] }
```
rekby: `test-util` is only needed by the new `#[tokio::test(start_paused = true)]` tests (paused clock, `advance`), but adding it to `[dependencies]` compiles it into the release `slo-native-topic` / `slo-native-query` binaries too. Consider moving just this feature to a `[dev-dependencies]` entry so it stays out of the shipped binaries — Cargo only applies dev-dependency features to test/bench/example builds, not to the binary crates that depend on `slo-framework`:

```toml
[dependencies]
tokio = { version = "=1.38.1", features = ["full", "signal"] }

[dev-dependencies]
tokio = { version = "=1.38.1", features = ["test-util"] }
```

rekby (commented):
Solid, well-tested change. The latency snapshot/reset module, the pending-span accounting (guarded by `pending` + `Drop`), the fair `RateLimiter`, and the worker supervision all look correct, and the metric names line up with `metrics-topic.yaml`. One thing worth a second look: the reader and writer workers treat SDK errors asymmetrically (reader aborts the whole workload, writer only records and continues) — see the inline comment. Everything else is minor.

rekby (commented):
OK

rekby (commented):
OK

rekby (commented):
OK

rekby (commented):
OK

rekby (commented):
OK

rekby (approved):

PR #548, opened 2026-07-17
feat: Rust SDK topic service benchmark

aral01:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There is no Rust SDK services benchmarks, therefore it is not possible to compare this SDK with C++ SDK.

Issue Number: N/A

## What is the new behavior?
This PR adds Rust implementation of topic service benchmark. 

Accepts config via JSON, returns result of benchmark as JSON. 
Both Rust and C++ have convenient mapping JSON <-> structs. 

Benchmark collects:
- Write acknowledgement latency
- End-to-end delivery latency
- Reader commit acknowledgement latency
- Read/write throughput 

Benchmark does:
- Runs concurrently configured readers/writers
- During warmup they just send/accept messages
- During measurement they send/accept messages and record latency/throughput statistics
- Then measuring ends, all tasks end, their statistics are merged and returned as result of benchmark

### DIFF
--- a/ydb/Cargo.toml
+++ b/ydb/Cargo.toml
@@ -63,12 +63,14 @@ flate2 = "1.1.9"
 rayon = "1.10"
 
 [dev-dependencies]
+anyhow = "1"
 async_once = "0.2"
 chrono = { version = "0.4", default-features = false, features = [
     "clock",
     "std",
 ] }
 lazy_static = "1.4"
+hdrhistogram = "7.5.4"
 ntest = "0.7"
 opentelemetry = "0.32.0"
 opentelemetry-otlp = { version = "0.32.0", features = ["grpc-tonic"] }
--- a/ydb/Cargo.toml
+++ b/ydb/Cargo.toml
@@ -78,3 +78,9 @@ opentelemetry_sdk = "0.32.1"
 tokio = { version = "1.22", features = ["test-util"] }
 tracing-opentelemetry = "0.33.0"
 uuid = { version = "1.17.0", features = ["v4"] }
+
+[[bench]]
+name = "sdk_compare"
+path = "benches/sdk_compare/main.rs"
+harness = false
+test = false
--- a/ydb/benches/sdk_compare/topic/reader.rs
+++ b/ydb/benches/sdk_compare/topic/reader.rs
@@ -1,0 +1,149 @@
+use std::time::{Duration, Instant};
+
+use anyhow::{Context, Result};
+use futures_util::stream::{FuturesUnordered, StreamExt};
+use tokio::task::JoinSet;
+use tokio::time::sleep_until;
+use ydb::{TopicClient, TopicReader, TopicReaderMessage, TopicReaderOptions, TopicSelector};
+
+use super::BenchmarkSchedule;
+use crate::config::TopicWorkload;
+use crate::metrics::LatencyRecorder;
+use crate::payload;
+
+pub(super) struct ReaderMetrics {
+    pub(super) end_to_end: LatencyRecorder,
+    pub(super) commit_ack: LatencyRecorder,
+}
+
+impl ReaderMetrics {
+    fn new() -> Result<Self> {
+        Ok(Self {
+            end_to_end: LatencyRecorder::new()?,
+            commit_ack: LatencyRecorder::new()?,
+        })
+    }
+
+    fn merge(&mut self, other: &Self) -> Result<()> {
+        self.end_to_end.merge(&other.end_to_end)?;
+        self.commit_ack.merge(&other.commit_ack)?;
+        Ok(())
+    }
+
+    fn record_commit(&mut self, latency: Option<Duration>) -> Result<()> {
+        if let Some(latency) = latency {
+            self.commit_ack.record(latency)?;
+        }
+        Ok(())
+    }
+}
+
+pub(super) async fn open(
+    topic_client: &mut TopicClient,
+    topic_path: &str,
+    workload: &TopicWorkload,
+) -> Result<Vec<TopicReader>> {
+    let mut readers = Vec::with_capacity(workload.reader_count);
+    for reader_id in 0..workload.reader_count {
+        let options = TopicReaderOptions::builder()
+            .consumer(workload.consumer_name.clone())
+            .topic(TopicSelector::new(topic_path))
+            .build();
+        let reader = topic_client
+            .create_reader_with_params(options)
+            .await
+            .with_context(|| format!("failed to create reader {reader_id}"))?;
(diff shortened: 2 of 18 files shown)


rekby (commented):
Reviewed the new `sdk_compare` topic benchmark end to end against the README contract and the SDK it drives. The code is careful and idiomatic: `anyhow`-based error propagation throughout, no `unwrap`/`expect`/`panic!`/`unsafe`, deadline-based draining via `timeout_at`, and warmup/measurement/drain boundaries handled symmetrically on the writer (`is_measurement_instant`) and reader (`sent_at_ns >= measurement_start_ns`) sides.

I verified two things that are easy to get wrong here:
- The `reader.read_batch()` call inside the `tokio::select!` loop is cancel-safe: `RuntimeHandle::pop_batch` only awaits when the buffer is empty and pops synchronously under the mutex, so losing the `select!` race drops no messages.
- The SDK APIs used compile as called from an external crate — `write_with_ack_future` returns an owned `AckFuture` (no borrow of the writer, so keeping acks `in_flight` while later calling `writer.stop()` is sound), `commit_with_ack` is `use<>`-owned, and `TopicReaderBatch.messages` is `pub`.

One caveat, not a blocker: the `[[bench]]` target is declared with `test = false`, so `cargo test --workspace` won't compile it. It's only built by `cargo bench` / `cargo clippy --all-targets`. If CI doesn't run one of those, this benchmark can silently bit-rot when the topic reader/writer API changes — worth confirming a CI job builds it.

rekby (commented):
OK

rekby (commented):
Reviewed the topic benchmark and the two production SDK changes it depends on.

The production-code changes are behavior-preserving and safe: `commit_with_ack` still calls `runtime.commit()` eagerly (only the future representation changed), and the `AckFuture` -> `TopicWriterAckFuture` rename is not an external break because `client_topic` is `pub(crate)` and the old name was never re-exported. No stray references to the old name remain.

The benchmark itself is carefully implemented: the per-writer semaphore + bounded channel keep in-flight submissions within `max_in_flight`, warmup/measurement attribution is symmetric between the writer and reader sides, and warm-up/measurement/drain scheduling is bounded by `completion_deadline`. One inherent (and already documented) property to keep in mind when comparing SDKs: readers stop reading at `measurement_end`, so messages written just before the boundary are never delivered and are excluded from `end_to_end`/read throughput. This biases read counts below write counts near the boundary; it's fine as long as comparisons only use identical scenarios, which the README calls out.

No blocking issues found.

ydb/Cargo.toml
```diff
@@ -63,16 +63,24 @@ flate2 = "1.1.9"
...
 
 [dev-dependencies]
+anyhow = "1"
 async_once = "0.2"
 chrono = { version = "0.4", default-features = false, features = [
     "clock",
     "std",
 ] }
 lazy_static = "1.4"
+hdrhistogram = "7.5.4"
```
rekby, line 73: Nit: the `[dev-dependencies]` list is otherwise alphabetical (`anyhow`, `async_once`, `chrono`, `lazy_static`, `ntest`, `opentelemetry`...), but `hdrhistogram` landed after `lazy_static`. Consider moving it up between `chrono` and `lazy_static` to keep the ordering consistent.

rekby (commented):
Reviewed the full diff and the delta. The benchmark is carefully written: the per-writer `Semaphore` + bounded `mpsc` give correct in-flight back-pressure (channel capacity equals permit count, so `try_send` cannot spuriously fail), a single monotonic `BenchmarkSchedule::origin` is shared by readers and writers so `end_to_end` latencies are consistent, warm-up/measurement partitioning via the payload timestamp is coherent on both sides, and all fallible paths propagate through `anyhow` with no `unwrap`/`expect`/`panic!` in the workload code. The SDK-side `commit_with_ack` refactor to a named `TopicReaderCommitAckFuture` preserves the previous error semantics (eager commit submission, closed-channel -> `YdbError::custom`) and adds tests, and the removed `RayonExecutor` test-only re-export is unused elsewhere. Only a small nit below.

ydb/benches/sdk_compare/topic/reader.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        anyhow::Ok(ack_latency)
+    });
+
+    let mut end_to_end = LatencyRecorder::new()?;
+
+    loop {
+        // Keep reading while completed commit acknowledgements are recorded in parallel.
+        let batch = tokio::select! {
```
rekby, line 108: tokio timeout

ydb/benches/sdk_compare/topic/reader.rs
```diff
@@ -0,0 +1,177 @@
...
+        .await
+        .context("commit acknowledgement recorder panicked or was cancelled")??;
+
+    Ok(ReaderMetrics {
+        end_to_end,
+        commit_ack,
+    })
+}
+
+async fn record_end_to_end_latencies(
```
rekby, line 152: разделить на отдельные функции - определения нужно ли мерять сообщения и отдельно - запись метрик.

ydb/benches/sdk_compare/topic/writer.rs
```diff
@@ -0,0 +1,197 @@
...
+        anyhow::Ok(ack_latency)
+    });
+
+    // Keep one continuous pipeline across warm-up and measurement.
+    loop {
+        let concurrency_permit = tokio::select! {
+            permit = Arc::clone(&write_slots).acquire_owned() => {
+                permit.context("write concurrency limiter was closed")?
+            }
+            () = sleep_until(schedule.measurement_end.into()) => break,
```
rekby, line 148: timeout

rekby (commented):
и переносим бенчмарки в отдельную репу

rekby (commented):
OK

ydb/benches/sdk_compare/topic/reader.rs
aral01: С таймаутом появляется грязь: бесконечный цикл раст выводит как Future<Output = !>, т.е. которая никогда не завершится. В ней нельзя использовать `?`. Для вывода типа необходимо оборачивать это в функцию Result Infallible. Потом по сигнатуре таймаут -- ошибка, которую необходимо проглотить, что опять в свою очередь будет вызывать вопросы при ревью. В предложенном варианте все батчи, полученные до дедлайна, будет до конца закомиченны и учтены в статистике

aral01:
Moved to: https://github.com/ydb-platform/ydb-sdk-benchmarks/pull/1

PR #549, opened 2026-07-17
fix: access token should be renewed before the first discovery attempt

zeromem88:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

`TimerDiscovery` starts making `ListEndpoints` gRPC calls in a background task immediately during `ClientBuilder::client()`, before the token has been obtained from the credentials provider. The first discovery calls pass through `AuthGrpcInterceptor`, which reads an empty token string (`""`) from `TokenCache` (initialised before the first `create_token()` call completes) and sends it as the `x-ydb-auth-ticket` header. The server rejects these unauthenticated requests, and the discovery retry loop repeats this until the token refresh finishes asynchronously in a separate background thread.

Issue Number: N/A

## What is the new behavior?

- `DiscoverySharedState` now holds a `token_waiter` — a `Box<dyn Waiter>` (the `TokenCache` itself, which implements `Waiter`).
- `background_discovery` waits for the token to be available before entering the discovery retry loop. The waiter is consumed once, so subsequent periodic rediscoveries are not affected.
- `ClientBuilder::client()` passes a clone of `db_cred.token_cache` as the waiter to `TimerDiscovery::new`.
- The first `ListEndpoints` call is guaranteed to carry a valid `x-ydb-auth-ticket` header.

## Other information

The `Client::wait()` method already calls `self.credentials.token_cache.wait().await?` before `self.discovery.wait().await?`, but the background discovery task is spawned during construction, long before the user calls `wait()`. This fix ensures the background task itself holds off, making the startup sequence deterministic regardless of token acquisition latency.

### DIFF
--- a/ydb/src/discovery.rs
+++ b/ydb/src/discovery.rs
@@ -12,7 +12,7 @@ use http::uri::Authority;
 use itertools::Itertools;
 use tokio::sync::watch;
 use tokio_stream::wrappers::WatchStream;
-use tracing::{instrument, trace};
+use tracing::{instrument, trace, warn};
 
 use crate::YdbError;
 use crate::errors::{NeedRetry, YdbResult};
@@ -207,17 +207,15 @@ impl TimerDiscovery {
         connection_manager: DiscoveryConnectionManager,
         endpoint: &str,
         interval: Duration,
-        token_waiter: Box<dyn Waiter>,
+        token_waiter: impl Waiter + 'static,
     ) -> YdbResult<Self> {
-        let state = Arc::new(DiscoverySharedState::new(
-            connection_manager,
-            endpoint,
-            token_waiter,
-        )?);
+        let state = Arc::new(DiscoverySharedState::new(connection_manager, endpoint)?);
 
         let state_weak = Arc::downgrade(&state);
         tokio::spawn(DiscoverySharedState::background_discovery(
-            state_weak, interval,
+            state_weak,
+            interval,
+            token_waiter,
         ));
 
         Ok(TimerDiscovery { state })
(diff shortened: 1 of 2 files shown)


ydb/src/discovery.rs
```diff
@@ -275,6 +279,11 @@ struct DiscoverySharedState {
 
     discovery_lock: tokio::sync::Mutex<()>,
 
+    /// Waiter for token readiness. Consumed on the first successful wait
+    /// inside `background_discovery`. Not used after that.
+    #[derivative(Debug = "ignore")]
+    token_waiter: tokio::sync::Mutex<Option<Box<dyn Waiter>>>,
+
```
liferooter: No need to store the waiter inside the state. It's easier to just wait for it in spawned task before starting background discovery.

ydb/src/discovery.rs
liferooter, line 219:
```suggestion
tokio::spawn(async move {
    token_waiter.wait().await.unwrap_or_else(|err| {
        trace!("token waiter returned error (ignored): {err}");
    });
    DiscoverySharedState::background_discovery(
        state_weak, interval,
    ).await;
});
```

ydb/src/discovery.rs
```diff
@@ -204,13 +204,17 @@ pub(crate) struct TimerDiscovery {
 }
 
 impl TimerDiscovery {
-    #[allow(dead_code)]
     pub(crate) fn new(
         connection_manager: GrpcConnectionManager,
         endpoint: &str,
         interval: Duration,
+        token_waiter: Box<dyn Waiter>,
```
liferooter: When the waiter is not stored explicitly, its type can be made `impl Waiter + 'static`, no need for boxing.

ydb/src/discovery.rs
zeromem88: Thanks! 
Removed waiter from the state.

ydb/src/discovery.rs
zeromem88: Fixed

ydb/src/discovery.rs
zeromem88: Fixed

rekby (commented):
OK

rekby (approved):

PR #550, opened 2026-07-20
Fix all found typos

liferooter:
Found and fixed a lot of typos with `typos`.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

There are a lot of typos in the code.

Issue Number: N/A

## What is the new behavior?

There are much less typos in the code.


rekby (commented):
OK

rekby (changes requested):
Тут нужно добавить линтер в CI и описать как его применять локально, чтобы ошибки больше не пролезали в репозиторий.

ydb/src/client_scheme/list_types.rs
```diff
@@ -15,7 +15,7 @@ pub enum SchemeEntryType {
     Unspecified,
     Directory,
     Table,
-    ColumnStrore,
+    ColumnStore,
```
rekby, line 18: `SchemeEntryType` is part of the public API (`pub use client_scheme::list_types::{.., SchemeEntryType, ..}` in `ydb/src/lib.rs:192`), so renaming this variant `ColumnStrore` → `ColumnStore` is a breaking change for any consumer that matches on the old name. The rename is the right thing to do, but it should land in a semver-major bump (and ideally be called out in release notes). Just flagging so it isn't shipped as a patch.

rekby (commented):
Clean, well-scoped typo sweep. I verified every renamed identifier is updated at all call sites — grepping the tree for the old spellings (`credencials`, `dicovery`, `ColumnStrore`, `macroses`, `succesfull`, `proccess`, `reciever`, `gurantee`, etc.) returns no matches, so nothing is left dangling and the crate should still compile. Internal renames (`credentials_ref`, the `discovery_pessimization_interceptor`/`macros` module & file renames, and the private `NearestDCBalancer`/`WaiterImpl` fields) are all `pub(crate)`/private and safe.

One thing worth a maintainer's eye: `SchemeEntryType` is re-exported from the crate root (`ydb/src/lib.rs:192`), so renaming its `ColumnStrore` variant is technically a semver-breaking change for any downstream code that matches on the old (misspelled) name. See the inline note — the fix itself is correct, just flagging the API impact.

Note: the large topic-reader refactor visible in the delta diff is not part of this PR's net changes (it came in via the `master` merge) — I reviewed only the typo/CI changes that actually belong to this PR.

rekby (approved):

ydb/src/client_scheme/list_types.rs
liferooter: Yes. But we can break public API for now.

PR #551, opened 2026-07-20
Add nightly to CI for codecov

liferooter:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

The crate is not tested on nightly compiler, also doctests are not included in coverage reports. Also being broken with the latest
versions of deps don't block merging.

Issue Number: N/A

## What is the new behavior?

The crate is tested on nightly compiler too, covecoverage tests are run only against nightly compiler because it support
including doctests in coverage tests, also being broken with the latest versions of deps now block merging as it should.

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -2,11 +2,11 @@ name: YDB tests
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
   schedule:
-    - cron: '18 4 * * *'
+    - cron: "18 4 * * *"
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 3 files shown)


.github/workflows/rust-tests.yml
```diff
@@ -2,39 +2,48 @@ name: YDB tests
...
-  RUST_VERSION_MSRV: "1.88"
-  # Pinned modern Rust version used for forward compatibility coverage.
-  RUST_VERSION_MODERN: "1.96"
-  # Stable Rust version used against dependencies test
-  RUST_VERSION_STABLE: stable
 
 jobs:
   tests:
-    name: ${{ matrix.name }}
+    name: tests (${{ case(matrix.label != '', matrix.label, format('{0} rustc', matrix.rust_version)) }})
```
rekby, line 17: `case(...)` is not a GitHub Actions expression function. The supported functions are `contains`, `startsWith`, `endsWith`, `format`, `join`, `toJSON`, `fromJSON`, `hashFiles` (plus the status checks `success`/`always`/`cancelled`/`failure`) — there is no `case`. GitHub will reject this expression ("Unrecognized function: 'case'"), which breaks name evaluation for the whole `tests` job.

Use the standard `&& / ||` ternary idiom instead — it works here because `matrix.label` is a non-empty (truthy) string whenever it's set, and the `stable` entry (which has no `label`) falls back to `stable rustc`:

```suggestion
    name: tests (${{ matrix.label != '' && matrix.label || format('{0} rustc', matrix.rust_version) }})
```

rekby (commented):
The intent (nightly leg for coverage incl. doctests, dropping `continue-on-error` so the latest-deps leg blocks merges) reads clearly and the doctest `recursion_limit` additions look right. One blocking problem: the job name uses a `case(...)` expression that isn't a real GitHub Actions function, so the `tests` job will fail to evaluate its name. Detail inline.

.github/workflows/rust-tests.yml
liferooter: Case _is_ a GitHub Actions expression function. Please go read GitHub docs.

rekby (commented):
OK

PR #566, opened 2026-07-24
fix: add cancellation safety checks

### LINKED ISSUE #563: dev: make TopicWriter `writer` cancellation safe
**Impact Assessment**

How would this change affect:
- [ ] Performance
- [ ] Maintainability
- [ ] Backward compatibility
- [x] User experience

**Area of Improvement**
Topic Writer cancellation safety.

**Current Situation**
`TopicWriter::write` function is not cancel safe.
In `Reconnector::add_message` happens given sequence of actions:
```rust
        let mut state_guard = self.state.lock().await;
        state_guard.connection_info.last_seq_no_assigned = message_seq_no;
        self.queue.add_message(message, ack_sender).await
```
If `write` is cancelled before message actually put in queue, but inner state of `last_seq_no_assigned` already mutated, writer will return errors on next message writes because of seq_no monotonicity violation.

**Suggested Improvement**

aral01:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): tests

Topic writer is cancel safe.

Issue Number: closes #563

## What is the new behavior?

- Test on cancellation safety on topic write with custom seq-no
- Test on cancellation safety on topic write with auto seq-no

### DIFF
--- a/ydb/src/client_topic/topicwriter/reconnector.rs
+++ b/ydb/src/client_topic/topicwriter/reconnector.rs
@@ -475,3 +475,86 @@ impl ReconnectionLoop {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::{sync::Arc, time::Duration};
+
+    use tokio::sync::{Mutex, watch};
+    use tokio_util::sync::CancellationToken;
+
+    use crate::{
+        TopicWriterMessage,
+        client_topic::topicwriter::{
+            connection::ConnectionInfo,
+            queue::Queue,
+            reconnector::{Reconnector, ReconnectorState, ReconnectorStatus},
+        },
+        grpc_wrapper::raw_topic_service::common::codecs::RawSupportedCodecs,
+    };
+
+    fn reconnector(auto_seq_no: bool) -> Reconnector {
+        let (_, rx) = watch::channel(ReconnectorStatus::Working);
+
+        Reconnector {
+            state: Arc::new(Mutex::new(ReconnectorState {
+                connection_info: ConnectionInfo {
+                    partition_id: 0,
+                    session_id: "".to_string(),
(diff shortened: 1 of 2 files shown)


rekby (commented):
Tests-only change verifying `Reconnector::add_message` is cancellation-safe (issue #563). Traced both paths: `add_message` mutates `last_seq_no_assigned` under the `state` lock and then awaits `queue.add_message`, which blocks on the inner queue lock held by the test — so the `biased` select polls the future once, finds it pending at exactly the queue-add await (before any message/reception ticket is enqueued), and drops it. The follow-up `add_message` then succeeds, correctly asserting no lock leak, poisoned state, or dangling ticket. Both custom and auto seq-no cases are deterministic (no timers or racing tasks) and match the stated intent. Looks correct; only a small readability nit inline.

ydb/src/client_topic/topicwriter/reconnector.rs
```diff
@@ -475,3 +475,86 @@ impl ReconnectionLoop {
...
+
+    #[tokio::test]
+    async fn add_message_cancellation_safety_custom_seq_no() {
+        let reconnector = reconnector(false);
+
+        let message = TopicWriterMessage::builder().data(vec![]).seq_no(2).build();
+
+        let queue_lock = reconnector.queue.lock().await;
+
+        tokio::select! {biased;
```
rekby: Optional readability nit: the `tokio::select! { biased; result = add_message => panic!(...), _ = async {} => {} }` construct is a fairly obscure way to say "poll `add_message` exactly once and assert it is still pending (blocked on the queue lock we hold)". Since the same idiom recurs in both tests, a one-line comment explaining the intent would help future readers. Not blocking.

rekby (commented):
OK

PR #568, opened 2026-07-28
Fix #267: add basic metrics for Prometheus

### LINKED ISSUE #267: feat: Prometheus
created automatically from https://ydb.tech/docs/en/reference/ydb-sdk/feature-parity

sergey-glyanets:
Add basic Prometheus metrics 

## Pull request type

- [ ] Bugfix
- [V] Feature
- [ ] Build related changes

Add basic Prometheus metrics: collect metrics

Issue Number: 267

## What is the new behaviour?

- add metrics' names storage
- add counters: new clients (query, scheme, topic), query execution (exec, commit, rollback) 
- add histogram: query execution time

### DIFF
--- a/ydb/Cargo.toml
+++ b/ydb/Cargo.toml
@@ -64,8 +64,11 @@ uuid = { version = "1.17.0", features = ["v4", "v7"] }
 ydb-grpc = { version = "0.2.2", path = "../ydb-grpc" }
 flate2 = "1.1.9"
 rayon = "1.10"
+metrics = "0.24.6"
 
 [dev-dependencies]
+prometheus = "0.14.0"
+metrics-prometheus = "0.11.2"
 async_once = "0.2"
 chrono = { version = "0.4", default-features = false, features = [
     "clock",
--- a/ydb/src/client_builder.rs
+++ b/ydb/src/client_builder.rs
@@ -1,4 +1,5 @@
 use crate::client_common::{DBCredentials, TokenCache};
+use crate::client_metrics::names::MetricsNames;
 use crate::client_topic::compression::Executor;
 use crate::credentials::{
     AccessTokenCredentials, CredentialsRef, GCEMetadata, ServiceAccountCredentials,
@@ -235,6 +236,7 @@ pub struct ClientBuilder {
     grpc_opts: GrpcOptions,
     executor: Option<Arc<dyn Executor>>,
     retry_settings: Option<RetrySettings>,
+    driver_name: Option<String>,
 }
 
 impl ClientBuilder {
@@ -271,6 +273,10 @@ impl ClientBuilder {
     }
 
     pub async fn build(self) -> YdbResult<Client> {
+        let metrics_names = MetricsNames::new(self.driver_name);
+
+        metrics_names.client_new_counter.increment(1);
+
         let retry_settings = self
             .retry_settings
             .unwrap_or_else(RetrySettings::with_default_backoff);
@@ -323,6 +329,7 @@ impl ClientBuilder {
             load_balancer,
             self.executor,
             retry_settings,
+            metrics_names,
         )
         .await
     }
@@ -376,6 +383,13 @@ impl ClientBuilder {
         self
     }
 
+    /// Set metrics driver name
+    /// If unset, `main` is used.
+    pub fn with_driver_name(mut self, driver_name: &str) -> Self {
+        self.driver_name = Some(driver_name.to_string());
+        self
+    }
+
     fn new() -> Self {
         Self {
             credentials: credentials_ref(AccessTokenCredentials::from("")),
@@ -387,6 +401,7 @@ impl ClientBuilder {
             grpc_opts: GrpcOptions::default(),
             executor: None,
             retry_settings: None,
+            driver_name: None,
         }
     }
 }
--- a/ydb/src/client_metrics/names.rs
+++ b/ydb/src/client_metrics/names.rs
@@ -1,0 +1,50 @@
+use std::string::ToString;
+
+use metrics::{Counter, Histogram, counter, histogram};
+
+const DEFAULT_DRIVER_NAME: &str = "main";
+
+#[derive(Clone, Debug)]
+pub(crate) struct MetricsNames {
+    pub client_new_counter: Counter,
+    pub client_new_table_client_counter: Counter,
+    pub client_new_query_client_counter: Counter,
+    pub client_new_scheme_client_counter: Counter,
+    pub client_new_topic_client_counter: Counter,
+    pub client_query_row_counter: Counter,
+    pub client_transaction_query_row_counter: Counter,
+    pub client_transaction_exec_counter: Counter,
+    pub client_transaction_commit_counter: Counter,
+    pub client_transaction_rollback_counter: Counter,
+    pub client_row_query_time_histogram: Histogram,
+    pub client_transaction_row_query_time_histogram: Histogram,
+}
+
+impl Default for MetricsNames {
+    fn default() -> Self {
+        MetricsNames::new(None)
+    }
+}
+
+impl MetricsNames {
+    pub fn new(driver_name: Option<String>) -> Self {
+        let labels = [(
+            "driver_name",
+            driver_name.unwrap_or(DEFAULT_DRIVER_NAME.to_string()),
+        )];
+        Self {
+            client_new_counter: counter!(description: "ydb new client counter", "ydb_new_client_counter", &labels),
+            client_new_table_client_counter: counter!(description: "ydb new table client counter", "ydb_new_table_client_counter", &labels),
+            client_new_query_client_counter: counter!(description: "ydb new query client counter", "ydb_new_query_client_counter", &labels),
+            client_new_scheme_client_counter: counter!(description: "ydb new scheme client counter", "ydb_new_scheme_client_counter", &labels),
+            client_new_topic_client_counter: counter!(description: "ydb new topic client counter", "ydb_new_topic_client_counter", &labels),
+            client_query_row_counter: counter!(description: "ydb client query row counter", "ydb_client_query_row_counter", &labels),
+            client_transaction_query_row_counter: counter!(description: "ydb client transaction query row counter", "ydb_client_transaction_query_row_counter", &labels),
+            client_transaction_exec_counter: counter!(description: "ydb client transaction exec counter", "ydb_client_transaction_exec_counter", &labels),
+            client_transaction_commit_counter: counter!(description: "ydb client transaction commit counter", "ydb_client_transaction_commit_counter", &labels),
+            client_transaction_rollback_counter: counter!(description: "ydb client transaction rollback counter", "ydb_client_transaction_rollback_counter", &labels),
+            client_row_query_time_histogram: histogram!(description: "ydb row query time histogram", "ydb_row_query_time_histogram", &labels),
+            client_transaction_row_query_time_histogram: histogram!(description: "ydb transaction row query time histogram", "ydb_transaction_row_query_time_histogram", &labels),
+        }
+    }
+}
--- a/ydb/src/client_query/builders.rs
+++ b/ydb/src/client_query/builders.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::future::IntoFuture;
 use std::marker::PhantomData;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use crate::TxMode;
 use crate::errors::{YdbError, YdbResult};
@@ -211,10 +211,22 @@ impl<'a, T: FromYdbRow + 'a, S> IntoFuture for CallBuilder<'a, OneRow<T>, S> {
 
     fn into_future(mut self) -> Self::IntoFuture {
         Box::pin(async move {
+            let start = Instant::now();
             let set = exactly_one_set(
                 materialize_query(&mut self.core, self.text, self.params, self.opts).await?,
             )?;
             let row = take_single_row(set)?.ok_or(YdbError::NoRows)?;
+            let delta = start.elapsed();
+            match &self.core {
+                ExecCoreRef::Client(core) => core
+                    .metrics_names
+                    .client_row_query_time_histogram
+                    .record(delta.as_secs_f64()),
+                ExecCoreRef::Transaction(core) => core
+                    .metrics_names
+                    .client_transaction_row_query_time_histogram
+                    .record(delta.as_secs_f64()),
+            }
             T::from_row(row)
         })
     }
@@ -226,10 +238,23 @@ impl<'a, T: FromYdbRow + 'a, S> IntoFuture for CallBuilder<'a, OptionalRow<T>, S
 
     fn into_future(mut self) -> Self::IntoFuture {
         Box::pin(async move {
+            let start = Instant::now();
             let set = exactly_one_set(
                 materialize_query(&mut self.core, self.text, self.params, self.opts).await?,
             )?;
-            take_single_row(set)?.map(T::from_row).transpose()
+            let row = take_single_row(set)?.map(T::from_row).transpose();
+            let delta = start.elapsed();
+            match &self.core {
+                ExecCoreRef::Client(core) => core
+                    .metrics_names
+                    .client_row_query_time_histogram
+                    .record(delta.as_secs_f64()),
+                ExecCoreRef::Transaction(core) => core
+                    .metrics_names
+                    .client_transaction_row_query_time_histogram
+                    .record(delta.as_secs_f64()),
+            }
+            row
         })
     }
 }
--- a/ydb/src/client_query/mod.rs
+++ b/ydb/src/client_query/mod.rs
@@ -167,12 +167,14 @@ impl QueryClient {
         connection_manager: GrpcConnectionManager,
         session_pool: SessionPool,
         retry_settings: RetrySettings,
+        metrics_names: MetricsNames,
     ) -> Self {
         Self {
             ctx: ClientExecContext {
                 connection_manager,
                 session_pool,
                 retry_settings,
+                metrics_names,
             },
         }
     }
@@ -292,6 +294,7 @@ impl QueryClient {
                         client.ctx.session_pool.clone(),
                         options.clone(),
                         wall_timeout.map(|d| retry.start_time + d),
+                        client.ctx.metrics_names.clone(),
                     );
 
                     client.try_attempt_body(callback, tx, idempotency).await
@@ -384,6 +387,7 @@ impl QueryExecutor for QueryClient {
     }
 
     fn query_row(&mut self, text: impl Into<String>) -> QueryRowBuilder<'_, Row, Self::Scope> {
+        self.ctx.metrics_names.client_query_row_counter.increment(1);
         QueryClient::query_row(self, text)
     }
 }
@@ -400,6 +404,7 @@ impl Transaction {
         session_pool: SessionPool,
         options: TransactionOptions,
         retry_deadline: Option<Instant>,
+        metrics_names: MetricsNames,
     ) -> Self {
         Self {
             ctx: transaction_exec_context(
@@ -407,6 +412,7 @@ impl Transaction {
                 session_pool,
                 options,
                 retry_deadline,
+                metrics_names,
             ),
         }
     }
@@ -505,6 +511,10 @@ impl QueryExecutor for Transaction {
     type Scope = builders::Interactive;
 
     fn exec(&mut self, text: impl Into<String>) -> ExecBuilder<'_, Self::Scope> {
+        self.ctx
+            .metrics_names
+            .client_transaction_exec_counter
+            .increment(1);
         Transaction::exec(self, text)
     }
 
@@ -517,10 +527,15 @@ impl QueryExecutor for Transaction {
     }
 
     fn query_row(&mut self, text: impl Into<String>) -> QueryRowBuilder<'_, Row, Self::Scope> {
+        self.ctx
+            .metrics_names
+            .client_transaction_query_row_counter
+            .increment(1);
         Transaction::query_row(self, text)
     }
 }
 
+use crate::client_metrics::names::MetricsNames;
 pub use builders::{
     CallBuilder, ClientOneShot, ExecBuilder, ExecCall, Interactive, OneResultSet, OneRow,
     OptionalRow, OptionalRowBuilder, QueryExecutor, QueryRowBuilder, QueryStreamBuilder,
@@ -586,6 +601,7 @@ mod unit_tests {
             pool.clone(),
             TransactionOptions::default(),
             None,
+            MetricsNames::new(None),
         );
         tx.ctx.pooled_lease = Some(lease);
         exec::transaction_handle_query_error(
--- a/ydb/tests/query_metrics.rs
+++ b/ydb/tests/query_metrics.rs
@@ -1,0 +1,589 @@
+//! Mock-server tests for query client metrics (counters).
+//!
+//! These tests are separated from `query_tx.rs` because they
+//! share the global `prometheus::default_registry()` and would
+//! interfere with parallel runs within the same test binary.
+#![recursion_limit = "256"]
+
+mod mock_server;
+
+use std::sync::LazyLock;
+
+use ydb::{Client, ClientBuilder, QueryExecutor, Transaction, YdbResult, closure};
+use ydb_grpc::ydb_proto::query::{ExecuteQueryResponsePart, TransactionMeta};
+use ydb_grpc::ydb_proto::status_ids::StatusCode;
+use ydb_grpc::ydb_proto::{Column, ResultSet, Type, Value, r#type};
+
+use crate::mock_server::handler::{FromHandlerToService, Handler, Incoming, Reply};
+use crate::mock_server::query::{QUERY_TX_ID, QueryIncoming, QueryReply};
+use crate::mock_server::server::MockServer;
+
+/// Serializes metric tests that share the global prometheus registry.
+static METRICS_LOCK: LazyLock<tokio::sync::Mutex<()>> =
+    LazyLock::new(|| tokio::sync::Mutex::new(()));
+
+const DATABASE: &str = "/local";
+
+async fn make_client(server: &MockServer) -> YdbResult<Client> {
+    ClientBuilder::new_from_connection_string(format!(
+        "{}{DATABASE}?use_discovery=false",
+        server.endpoint()
+    ))?
+    .build()
+    .await
+}
+
+fn counter_value(metric_name: &str) -> u64 {
+    let gathered = prometheus::default_registry().gather();
+    gathered
+        .iter()
+        .find(|mf| mf.name() == metric_name)
+        .map(|mf| mf.metric.iter().map(|m| m.counter.value() as u64).sum())
+        .unwrap_or(0)
+}
+
+fn counter_label_value(metric_name: &str, label_key: &str) -> Option<String> {
+    let gathered = prometheus::default_registry().gather();
+    gathered
+        .iter()
+        .find(|mf| mf.name() == metric_name)
+        .and_then(|mf| mf.metric.first())
+        .and_then(|m| {
+            m.label
+                .iter()
+                .find(|l| l.name() == label_key)
+                .map(|l| l.value().to_string())
+        })
+}
+
+fn row_with_value(value: i64) -> ExecuteQueryResponsePart {
+    ExecuteQueryResponsePart {
+        status: StatusCode::Success as i32,
+        issues: vec![],
+        result_set_index: 0,
+        result_set: Some(ResultSet {
+            columns: vec![Column {
+                name: "val".to_string(),
+                r#type: Some(Type {
+                    r#type: Some(r#type::Type::TypeId(
+                        ydb_grpc::ydb_proto::r#type::PrimitiveTypeId::Int64 as i32,
+                    )),
+                }),
+            }],
+            rows: vec![Value {
+                items: vec![Value {
+                    value: Some(ydb_grpc::ydb_proto::value::Value::Int64Value(value)),
+                    ..Default::default()
+                }],
+                ..Default::default()
+            }],
+            ..Default::default()
+        }),
+        exec_stats: None,
+        tx_meta: None,
+    }
+}
+
+#[tokio::test]
+#[tracing_test::traced_test]
+async fn happy_path_collect_metrics() -> YdbResult<()> {
+    let _guard = METRICS_LOCK.lock().await;
+    let _ = metrics_prometheus::try_install().ok(); // here should be `ok` because only one error can be emitted - duplicate attempt
+
+    struct DummyHandler;
+    impl Handler for DummyHandler {}
+
+    let (server, _reply_tx) = MockServer::start(DummyHandler).await;
+    let _ = make_client(&server).await?;
+
+    let metrics_vec = prometheus::default_registry().gather();
+    assert!(!metrics_vec.is_empty());
+
+    Ok(())
+}
+
+#[tokio::test]
+#[tracing_test::traced_test]
+async fn query_client_creation_increments_counter_metric() -> YdbResult<()> {
+    let _guard = METRICS_LOCK.lock().await;
+    let _ = metrics_prometheus::try_install().ok(); // here should be `ok` because only one error can be emitted - duplicate attempt
+
+    struct DummyHandler;
+    impl Handler for DummyHandler {}
+
+    let (server, _reply_tx) = MockServer::start(DummyHandler).await;
+    let client = make_client(&server).await?;
+
+    let before = counter_value("ydb_new_query_client_counter");
+
+    let _qc1 = client.query_client();
+    assert_eq!(counter_value("ydb_new_query_client_counter") - before, 1);
+
+    let _qc2 = client.query_client();
+    assert_eq!(counter_value("ydb_new_query_client_counter") - before, 2);
+
+    Ok(())
+}
+
+#[tokio::test]
+#[tracing_test::traced_test]
+async fn query_client_creation_with_driver_name_label() -> YdbResult<()> {
+    let _guard = METRICS_LOCK.lock().await;
+    let _ = metrics_prometheus::try_install().ok();
+
+    struct DummyHandler;
+    impl Handler for DummyHandler {}
+
+    let (server, _reply_tx) = MockServer::start(DummyHandler).await;
+
+    let client = ClientBuilder::new_from_connection_string(format!(
+        "{}{}?use_discovery=false",
+        server.endpoint(),
+        DATABASE,
+    ))?
+    .with_driver_name("custom")
+    .build()
+    .await?;
+
+    let _qc = client.query_client();
+
+    let driver_name = counter_label_value("ydb_new_query_client_counter", "driver_name");
+    assert_eq!(
+        driver_name.as_deref(),
+        Some("custom"),
+        "query_client counter must carry the custom driver_name label"
+    );
+
+    Ok(())
+}
+
+#[tokio::test]
+#[tracing_test::traced_test]
+async fn query_row_increments_counter_metric() -> YdbResult<()> {
+    let _guard = METRICS_LOCK.lock().await;
+    let _ = metrics_prometheus::try_install().ok(); // here should be `ok` because only one error can be emitted - duplicate attempt
+
+    struct QueryRowHandler {
+        replies: FromHandlerToService,
+    }
+    impl Handler for QueryRowHandler {
+        fn set_channel(&mut self, tx: FromHandlerToService) {
+            self.replies = tx;
+        }
+        fn handle(&self, incoming: Incoming) -> Option<Incoming> {
+            let Incoming::Query(QueryIncoming::ExecuteQuery(_, stream_id)) = incoming else {
+                return Some(incoming);
+            };
+            self.replies
+                .send(Reply::Query(QueryReply::ExecuteQuery {
+                    stream_id,
+                    part: row_with_value(42),
+                }))
+                .expect("mock response channel must remain open");
+            self.replies
+                .send(Reply::Query(QueryReply::ExecuteQueryClose { stream_id }))
+                .expect("mock response channel must remain open");
+            None
+        }
+    }
+
+    let handler = QueryRowHandler {
+        replies: tokio::sync::mpsc::unbounded_channel::<Reply>().0,
+    };
+    let (server, _reply_tx) = MockServer::start(handler).await;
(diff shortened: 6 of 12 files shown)


ydb/src/client_query/mod.rs
```diff
@@ -384,6 +387,8 @@ impl QueryExecutor for QueryClient {
     }
 
     fn query_row(&mut self, text: impl Into<String>) -> QueryRowBuilder<'_, Row, Self::Scope> {
+        metrics::counter!(self.ctx.metrics_names.client_query_row_counter_name.clone())
+            .increment(1);
```
rekby: Double-count: `QueryClient::query_row` is generated by `impl_client_query_methods!` (see `builders.rs`), whose body already does `metrics::counter!(self.ctx.metrics_names.client_query_row_counter_name.clone()).increment(1)`. This trait impl increments the same counter and then delegates to that inherent method, so every call through the `QueryExecutor` trait counts twice.

The `Transaction` side counts only in the trait impl (its inherent macro method doesn't increment), so to stay consistent, drop the increment either here or in the macro — but not both.

rekby (commented):
Thanks for adding Prometheus metrics. Two correctness issues around the `query_row` counter make its values unreliable: the QueryClient path double-counts, and the Transaction path uses a hardcoded metric name that ignores the configurable prefix. Details inline.

ydb/src/client_query/mod.rs
```diff
@@ -529,10 +564,12 @@ impl QueryExecutor for Transaction {
     }
 
     fn query_row(&mut self, text: impl Into<String>) -> QueryRowBuilder<'_, Row, Self::Scope> {
+        metrics::counter!("ydb_client_query_row_count").increment(1);
```
rekby: This uses a hardcoded literal instead of the configured name from `metrics_names`. It ignores the `with_metrics_prefix` prefix and also diverges from the naming used everywhere else (`_query_row_counter` vs `_query_row_count`), so this counter won't be found under a custom prefix and splits from the intended metric.

```suggestion
        metrics::counter!(self.ctx.metrics_names.client_query_row_counter_name.clone()).increment(1);
```

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,65 @@
...
+
+impl MetricsNames {
+    pub fn new() -> Self {
+        let mut n: MetricsNames = Default::default();
+        n.set_prefix("ydb_client");
+        n
+    }
+    pub fn set_prefix(&mut self, metrics_prefix: &str) {
+        self.metrics_prefix = metrics_prefix.to_string();
+        self.client_new_counter_name = metrics_prefix.to_string().add("_new_counter").to_string();
```
rekby: Nit (readability): `metrics_prefix.to_string().add("_new_counter").to_string()` is quite roundabout — `metrics_prefix` is already `&str`, and the trailing `.to_string()` on an owned `String` is redundant. `format!` reads cleaner and the same applies to each assignment below (and lets you drop `use std::ops::Add;`).

```suggestion
        self.client_new_counter_name = format!("{metrics_prefix}_new_counter");
```

sergey-glyanets: Applied

ydb/src/client_query/mod.rs
sergey-glyanets: Fixed. Thanks!

ydb/src/client_builder.rs
```diff
@@ -251,6 +253,8 @@ impl ClientBuilder {
                 client_builder = handler(s, client_builder)?;
             }
         }
+        metrics::counter!(client_builder.metrics_names.client_new_counter_name.clone())
+            .increment(1);
```
rekby: `with_metrics_prefix` can only be called after this constructor returns, so this increment always uses the default `ydb_client_new_counter`. For example, `.with_metrics_prefix("foo").build()` never attributes the new-client count to `foo_new_counter`. Emit this metric after the prefix is configured, preferably when `build` successfully creates the client.

ydb/src/client_query/mod.rs
```diff
@@ -486,6 +514,13 @@ impl QueryExecutor for Transaction {
     type Scope = builders::Interactive;
 
     fn exec(&mut self, text: impl Into<String>) -> ExecBuilder<'_, Self::Scope> {
+        metrics::counter!(
+            self.ctx
+                .metrics_names
+                .client_transaction_exec_counter_name
+                .clone()
+        )
+        .increment(1);
```
rekby: These increments live only in the trait implementations, but `Transaction` has an inherent `exec` generated by `impl_transaction_query_methods!`. Rust resolves ordinary `tx.exec(...)` calls to that inherent method, so this counter remains zero for the normal SDK call path. The same applies to `Transaction::query_row` below and `QueryClient::query_row` above. Put the instrumentation in the generated inherent methods or a shared builder execution path, and let the trait implementations delegate without incrementing.

ydb/Cargo.toml
```diff
@@ -61,6 +61,9 @@ uuid = { version = "1.17.0", features = ["v4", "v7"] }
 ydb-grpc = { version = "0.2.2", path = "../ydb-grpc" }
 flate2 = "1.1.9"
 rayon = "1.10"
+metrics = "0.24.6"
+prometheus = "0.14.0"
+metrics-prometheus = "0.11.2"
```
rekby: These two crates are referenced only by `ydb/tests/query_tx.rs`; production code uses only the `metrics` facade. Keeping the Prometheus recorder and client in `[dependencies]` makes every SDK consumer build an otherwise unused exporter stack. Please move both to `[dev-dependencies]` unless the library itself is going to expose or use them at runtime.

ydb/tests/query_tx.rs
```diff
@@ -277,6 +277,21 @@ impl Handler for CommitTransportFailsHandler {
...
+#[tracing_test::traced_test]
+async fn happy_path_collect_metrics() -> YdbResult<()> {
+    let _ = metrics_prometheus::install();
+
+    let (handler, _tx_lifecycle) = CountingHandler::new();
+    let (server, _reply_tx) = MockServer::start(handler).await;
+    let _ = make_client(&server).await?;
+
+    let metrics_vec = prometheus::default_registry().gather();
+    assert_ne!(metrics_vec.len(), 0);
```
rekby: This only proves that some metric family exists after client construction. It never executes `query_row`, `exec`, commit, rollback, or the histogram path, so all advertised operation metrics can be missing or misnamed while the test stays green. Exercise the relevant operations and assert their exact metric names and values, including at least one custom-prefix case.

rekby (commented):

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub client_new_topic_client_counter_name: String,
+    pub client_query_row_counter_name: String,
+    pub client_transaction_query_row_counter_name: String,
+    pub client_transaction_exec_counter_name: String,
+    pub client_transaction_commit_counter_name: String,
+    pub client_transaction_rollback_counter_name: String,
+    pub client_row_query_time_histogram_name: String,
+}
+
+impl MetricsNames {
```
acacita: pls don't store them as strings to avoid repeating the metrics::counter!(name.clone()) everywhere. also hardcoded strings will be impossible to evolve or rename or regroup once the need for custom client-modified metrics occur.   

instead define the observability contract using a small internal facade package with methods for SDK-owned events. 

following comment applies to other code in this pr:
pls do not pollute every public method with counters histograms etc.
put metrics in raw gRPC wrapper calls, retry loop helpers,
session pool acquire/create/return/invalidate paths

acacita:
pls check out how they did it in go sdk: metrics are wrapped over instrumentation, not scattered across the whole codebase

rekby (commented):
OK

ydb/Cargo.toml
acacita, line 1: please make it feature gated
features = ["metrics"]

rekby (commented):
OK

ydb/Cargo.toml
sergey-glyanets: Thanks. Fixed.

ydb/src/client_builder.rs
sergey-glyanets: Moved. Thanks.

ydb/tests/query_tx.rs
```diff
@@ -277,6 +277,21 @@ impl Handler for CommitTransportFailsHandler {
     }
 }
 
+#[tokio::test]
+#[tracing_test::traced_test]
+async fn happy_path_collect_metrics() -> YdbResult<()> {
```
rekby: This discards the `Option` returned by `Result::ok()`. Since `Option` is `#[must_use]`, the repository's Clippy gate with `-D warnings` rejects this as `unused_must_use`; it also hides recorder installation failures. This test binary has only this installer, so require it to succeed.

```suggestion
    metrics_prometheus::try_install().expect("metrics recorder must install");
```

ydb/src/client_query/builders.rs
```diff
@@ -205,10 +207,23 @@ impl<'a, T: FromYdbRow + 'a, S> IntoFuture for CallBuilder<'a, OneRow<T>, S> {
...
+            match &self.core {
+                super::internal::ExecCoreRef::Client(core) => metrics::histogram!(
+                    core.metrics_names
+                        .client_row_query_time_histogram_name
+                        .clone()
+                )
+                .record(delta.as_secs_f64()),
+                super::internal::ExecCoreRef::Transaction(_) => {
+                    trace!("Transaction exec context metrics not yet implemented")
+                }
```
rekby: `Transaction::query_row` uses this same `IntoFuture`, but this branch records no latency, so transaction queries are absent from the advertised query-execution histogram. `TransactionExecContext` already carries `metrics_names`; record its histogram here as for the client branch, then remove the unused `trace` import.

rekby (commented):

ydb/src/client_builder.rs
rekby: The custom-prefix issue is fixed, but the increment still precedes several fallible steps and `Client::init(...).await`. A failed build therefore increments `*_new_counter` even though no `Client` was created. Preserve the metric name, await successful initialization, and increment afterward.

ydb/src/client_query/mod.rs
rekby: This still applies. The generated inherent methods remain uninstrumented, while only the trait implementations increment. The new test explicitly calls `QueryExecutor::query_row` via UFCS, so ordinary `qc.query_row(...)`, `tx.exec(...)`, and `tx.query_row(...)` still resolve to inherent methods and remain uncounted.

ydb/tests/query_tx.rs
rekby: Partially addressed, but the new tests still omit exact `exec`, commit, rollback, and histogram assertions. The `query_row` test also invokes the trait-qualified method, so it does not cover the ordinary inherent-method call path where the counter remains missing.

ydb/tests/query_metrics.rs
```diff
@@ -0,0 +1,280 @@
...
+        fn handle(&self, incoming: Incoming) -> Option<Incoming> {
+            let Incoming::Query(QueryIncoming::ExecuteQuery(_, stream_id)) = incoming else {
+                return Some(incoming);
+            };
+            self.replies
+                .send(Reply::Query(QueryReply::ExecuteQuery {
+                    stream_id,
+                    part: row_with_value(42),
+                }))
+                .ok();
```
rekby: `Option` is `#[must_use]`, so discarding the result of `.ok()` triggers `unused_must_use`; CI promotes that warning to an error. The identical calls on lines 167, 229, and 232 need the same fix. A closed mock response channel is a broken test invariant, so require the send to succeed.

```suggestion
                .expect("mock response channel must remain open");
```

rekby (commented):

ydb/tests/query_metrics.rs
sergey-glyanets: Fixed

rekby (commented):
OK

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,48 @@
+use std::string::ToString;
+
+#[derive(Default, Clone, Debug)]
+pub(crate) struct MetricsNames {
+    pub metrics_prefix: String,
```
rekby: Давай сделаем одну структуру сразу с объектами метрик

```rust
new_table_client_counter: Counter,
query_row_counter: Counter,
query_row_times: Histogram
```

и инициализируем их один раз, структуру можно копировать в новых клиентов - метрики легковесные.
Это позволит избежать клонирования строк и процедуры поиска объекта метрики на каждое изменение счётчика - код будет более читаемым.


И ещё подумал, что использование метки (label) для разделения экзепляров драйверов будет удобнее префикса. Давай оставим имена метрик константами, а вместо префикса добавим параметр .with_drive_name и будем использовать его как label driver_name, по-умолчанию: main.

Тогда легко будет строить разрезы одной метрики по разным драйверам - например latency запросов в разрезе по драйверу.

ydb/src/client_metrics/names.rs
rekby: Thanks, its good point about clone string for every metric increase.

About second part - about internal/public metrics: it will be next step. Main goal of the PR - create structure/internal API for metrics.

rekby (changes requested):

rekby (commented):
OK

ydb/src/client_metrics/names.rs
rekby: > also hardcoded strings will be impossible to evolve or rename or regroup once the need for custom client-modified metrics occur.

Can you describe why you need to rename and re-group metrics?

rekby (commented):
OK

ydb/src/client_metrics/names.rs
acacita: > Can you describe why you need to rename and re-group metrics?

Some examples from my experience. 

1. Preexisting organisation metric naming conventions
2. YDB has a beta version of OLAP tables and they suggest using separate databases for OLTP and OLAP. So one app may use multiple SDK clients in one process and will need for database="analytics" separation.
3. One app may have different workloads: e.g basic CRUD APIs, background jobs and migrations which need separate metrics views
4. Multi-tenant SDK health monitoring. Let's say we use YDB and one of their direct competitors and we want to monitor each SDK's health without running multiple processes for metrics scrapers/export paths 

anyways, I suggest using rust enums. also pls take note how the metrics are named according to prometheus naming conventions. 

also maybe some of suggested metrics would be nice to use in first iteration of this feature. rpc/tx total/session_pool are more crucial to SDK's health monitoring than just basic counters. 

```
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub(crate) enum MetricName {
    RpcRequests,
    RpcDuration,
    RetryAttempts,
    SessionPoolSessions,
    SessionPoolAcquireDuration,
    Transactions,
}

impl MetricName {
    pub(crate) const fn as_str(self) -> &'static str {
        match self {
            Self::RpcRequests => "rpc_requests_total",
            Self::RpcDuration => "rpc_duration_seconds",
            Self::RetryAttempts => "retry_attempts_total",
            Self::SessionPoolSessions => "session_pool_sessions",
            Self::SessionPoolAcquireDuration => "session_pool_acquire_duration_seconds",
            Self::Transactions => "transactions_total",
        }
    }
}

```

and for configuring something like

```

#[derive(Debug, Clone)]
pub(crate) struct MetricsConfig {
    namespace: Option<String>,
    static_labels: Vec<(&'static str, String)>,
}

impl Default for MetricsConfig {
    fn default() -> Self {
        Self {
            namespace: "ydb".to_string(),
            static_labels: Vec::new(),
        }
    }
}

```

Then the issues in points 2–4 are resolved, and we get stable ydb_* names by default, and can configure namespace if needed.

With renaming... well, it's totally optional, but may be useful as a possible evolvement path.

rekby (commented):
OK

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,69 @@
...
+            client_new_query_client_counter: counter!(description: "ydb new query client counter", "ydb_new_query_client_counter", &labels),
+            client_new_scheme_client_counter: counter!(description: "ydb new scheme client counter", "ydb_new_scheme_client_counter", &labels),
+            client_new_topic_client_counter: counter!(description: "ydb new topic client counter", "ydb_new_topic_client_counter", &labels),
+            client_query_row_counter: counter!(description: "ydb client query row counter", "ydb_client_query_row_counter", &labels),
+            client_transaction_query_row_counter_name: String::new(),
+            client_transaction_exec_counter_name: String::new(),
+            client_transaction_commit_counter_name: String::new(),
+            client_transaction_rollback_counter: counter!(description: "ydb client transaction rollback counter", "ydb_client_transaction_rollback_counter", &labels),
+            client_row_query_time_histogram: histogram!(description: "ydb new row query time histogram", "ydb_row_query_time_histogram", &labels),
```
rekby: This is a compile blocker: `metrics` 0.24's `counter!` and `histogram!` macros do not accept a `description:` argument. Register descriptions separately with `describe_counter!` / `describe_histogram!`, then create these handles with `counter!("…", &labels)` and `histogram!("…", &labels)`.

rekby (commented):

rekby (commented):
The new push leaves the previously reported compile blocker and three behavioral instrumentation issues unchanged; follow-ups are kept in their existing threads.

ydb/src/client_metrics/names.rs
rekby: Still present at the current head: `Default::default()` retains the unsupported `description:` arguments on lines 28–38, so `metrics` 0.24 cannot expand these macros.

ydb/src/client_builder.rs
rekby: Still unfixed: the increment precedes `AuthGrpcInterceptor::new`, discovery construction, and `Client::init(...).await`, all of which can fail. A failed build is therefore still counted as a new client.

ydb/src/client_query/mod.rs
rekby: Still unfixed: the macro-generated inherent methods only create builders, while the increments remain exclusively in the trait implementations. Ordinary `qc.query_row(...)`, `tx.exec(...)`, and `tx.query_row(...)` calls still bypass the counters.

ydb/src/client_query/builders.rs
rekby: Still unfixed: the transaction branch only emits a trace and never records `client_row_query_time_histogram`, even though its context now carries `metrics_names`.

ydb/src/client_metrics/names.rs
sergey-glyanets: Successfully compiled with metrics 0.24.6. Also, the [documentation](https://docs.rs/metrics/latest/metrics/macro.counter.html#named-parameters) for 0.24.6 version has a sample with "description" named argument.

rekby (commented):
OK

rekby (commented):

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,69 @@
...
+        self.client_new_counter = counter!("ydb_new_client_counter", &labels);
+        self.client_new_table_client_counter = counter!("ydb_new_table_client_counter", &labels);
+        self.client_new_query_client_counter = counter!("ydb_new_query_client_counter", &labels);
+        self.client_new_scheme_client_counter = counter!("ydb_new_scheme_client_counter", &labels);
+        self.client_new_topic_client_counter = counter!("ydb_new_topic_client_counter", &labels);
+        self.client_query_row_counter = counter!("ydb_client_query_row_counter", &labels);
+        self.client_transaction_query_row_counter_name =
+            format!("{driver_name}_transaction_query_row_counter");
+        self.client_transaction_exec_counter_name =
+            format!("{driver_name}_transaction_exec_counter");
```
rekby: These two transaction metrics still encode `driver_name` in the metric name and perform a fresh `counter!` lookup at every increment. Unlike the other metrics, clients named `main` and `analytics` therefore produce separate metric families instead of one stable `ydb_*` family partitioned by the `driver_name` label. Store `Counter` handles for fixed transaction metric names using `&labels`, as is already done for commit and rollback.

ydb/src/client_query/mod.rs
```diff
@@ -409,17 +415,29 @@ impl Transaction {
...
             return Ok(());
         }
         transaction_rollback(&mut self.ctx).await
     }
 
     async fn commit(&mut self) -> YdbResult<()> {
+        self.ctx
+            .metrics_names
+            .client_transaction_commit_counter
+            .increment(1);
```
rekby: This misses transactions committed through a final query with `.with_commit(true)`. That path is confirmed in `transaction_finish_committed_via_query` and returns from `resolve_post_callback_action` as already committed, so it never calls this method and never increments the commit counter. Instrument both confirmed commit paths (preferably at the shared state-transition layer).

ydb/src/client_query/mod.rs
```diff
@@ -409,17 +415,29 @@ impl Transaction {
...
         transaction_commit(&mut self.ctx).await
     }
 
     async fn rollback_quiet(&mut self) {
+        self.ctx
+            .metrics_names
+            .client_transaction_rollback_counter
+            .increment(1);
```
rekby: This increment is not tied to an actual rollback: `rollback_quiet` increments even when the state is already invalidated/finished and the following guard skips `transaction_rollback`. Conversely, real internal rollbacks such as the rollback inside `transaction_commit` after a hook failure and rollback-on-drop bypass this wrapper and are not counted. Place the metric on the actual rollback execution paths so the counter neither records no-ops nor misses real rollbacks.

ydb/tests/query_metrics.rs
```diff
@@ -0,0 +1,353 @@
...
+    let _ = metrics_prometheus::try_install().ok(); // here should be `ok` because only one error can be emitted - duplicate attempt
+
+    struct DummyHandler;
+    impl Handler for DummyHandler {}
+
+    let (server, _reply_tx) = MockServer::start(DummyHandler).await;
+    let _ = make_client(&server).await?;
+
+    let metrics_vec = prometheus::default_registry().gather();
+    assert_ne!(metrics_vec.len(), 0);
```
rekby: The workspace Clippy gate enables `-D warnings`, and `assert_ne!(collection.len(), 0)` triggers `clippy::len_zero`.

```suggestion
    assert!(!metrics_vec.is_empty());
```

ydb/src/client_metrics/names.rs
rekby: The `exec` half is fixed: it now uses the stored `client_transaction_exec_counter`. The query-row half remains unchanged: it still constructs `{driver_name}_transaction_query_row_counter` and calls `counter!` per invocation, so it still creates driver-specific metric families instead of using a fixed metric name with the `driver_name` label.

ydb/src/client_builder.rs
rekby: Still present after this push: `client_new_counter` is incremented at the start of `build`, before the fallible interceptor/discovery setup and `Client::init(...).await`.

ydb/src/client_query/mod.rs
rekby: Still present after this push. Only the transaction query-row handle lookup changed; the macro-generated inherent methods in `builders.rs` remain uninstrumented, so ordinary method calls still bypass these counters.

ydb/src/client_query/builders.rs
rekby: Still present after this push: the transaction arm remains a trace-only branch and does not record `client_row_query_time_histogram`.

ydb/src/client_query/mod.rs
rekby: Still present after this push: `transaction_finish_committed_via_query` transitions to `Committed` without incrementing the commit counter.

ydb/src/client_query/mod.rs
rekby: Still present after this push: `rollback_quiet` increments before checking whether the transaction is active, while internal and drop rollback paths still bypass this counter.

ydb/tests/query_metrics.rs
rekby: Still present after this push: `assert_ne!(metrics_vec.len(), 0)` remains unchanged and triggers the configured `clippy::len_zero` warning under `-D warnings`.

rekby (commented):
This push leaves the reviewed code unchanged, so the existing findings about failed client builds being counted, inherent query methods bypassing counters, missing transaction query latency, commit-via-query accounting, and rollback accounting still apply. No new findings.

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,69 @@
...
+            client_row_query_time_histogram: histogram!(description: "ydb new row query time histogram", "ydb_row_query_time_histogram", &labels),
+        }
+    }
+}
+
+impl MetricsNames {
+    pub fn new() -> Self {
+        let mut n: MetricsNames = Default::default();
+        n.set_driver_name(DEFAULT_DRIVER_NAME);
+        n
```
rekby: These macros return handles bound to the recorder active when they are called; without an installed recorder, the stored handles remain no-ops. Because `ClientBuilder::new()` calls `MetricsNames::new()` immediately, installing a recorder after constructing the builder—even before `build()` and every operation—silently loses all metrics. Defer handle creation until the recorder is expected to be installed, initialize lazily, or otherwise make this ordering requirement explicit and enforceable.

rekby (commented):

ydb/src/client_builder.rs
rekby: This remains unresolved: `build` increments at line 260, before fallible interceptor/discovery construction and `Client::init(...).await`, so failed builds are still counted as new clients.

ydb/src/client_query/mod.rs
rekby: The generated inherent methods still only construct `CallBuilder`s. Consequently ordinary `qc.query_row(...)`, `tx.exec(...)`, and `tx.query_row(...)` calls continue to bypass the increments that exist only in the trait implementations.

ydb/src/client_query/builders.rs
rekby: The transaction arm still emits only `trace!("Transaction exec context metrics not yet implemented")`; it does not record the histogram handle already present in `TransactionExecContext`.

ydb/src/client_query/mod.rs
rekby: `transaction_finish_committed_via_query` still transitions directly to `TxState::Committed` without incrementing the commit counter, so `.with_commit(true)` commits remain absent.

ydb/src/client_query/mod.rs
rekby: Moving the increment into `transaction_rollback` fixes the wrapper-only placement, but the metric is still incremented when there is no transaction ID/lease and therefore no rollback RPC. Conversely, `Drop` performs a rollback through `spawn_query_tx_rollback_on_drop` without incrementing it, so the counter still both overcounts no-op paths and misses real rollback requests.

rekby (commented):

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,69 @@
...
+            client_transaction_commit_counter: counter!(description: "ydb client transaction commit counter", "ydb_client_transaction_commit_counter", &labels),
+            client_transaction_rollback_counter: counter!(description: "ydb client transaction rollback counter", "ydb_client_transaction_rollback_counter", &labels),
+            client_row_query_time_histogram: histogram!(description: "ydb new row query time histogram", "ydb_row_query_time_histogram", &labels),
+        }
+    }
+}
+
+impl MetricsNames {
+    pub fn new() -> Self {
+        let mut n: MetricsNames = Default::default();
```
rekby: `MetricsNames::new()` registers every handle with `driver_name="main"` before the builder can apply `.with_driver_name(...)`. Metric registration creates that labeled series immediately, so an application that only builds an `analytics` driver still exports zero-valued `main` series for every metric, followed by the intended `analytics` series. Keep only the driver-name configuration in the builder and create the handles once after the final name is known; a custom-driver test should also assert that no `main` series exists.

ydb/src/client_metrics/names.rs
rekby: Still applicable: `ClientBuilder::new()` calls `MetricsNames::new()` at line 386, so all handles are bound before `build()`. The tests install the recorder before constructing the builder and therefore do not cover the failing order.

ydb/src/client_builder.rs
rekby: Still applicable: line 260 increments before the fallible interceptor/discovery setup and the awaited `Client::init`, so every failed build remains counted.

ydb/src/client_query/mod.rs
rekby: Still applicable: the inherent methods generated in `builders.rs` only construct `CallBuilder`s, while the increments remain solely in the trait methods at lines 359, 494, and 510.

ydb/src/client_query/builders.rs
rekby: Still applicable: the transaction arm at lines 227–229 remains trace-only and never records the histogram.

ydb/src/client_query/mod.rs
rekby: Still applicable: line 655 increments before checking for a transaction ID and lease, while the real rollback RPC in `spawn_query_tx_rollback_on_drop` still has no increment.

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,72 @@
+use std::string::ToString;
+
+use metrics::{Counter, Histogram, counter, histogram};
+
+const DEFAULT_DRIVER_NAME: &str = "main";
+
+#[derive(Clone, Debug)]
+pub(crate) struct MetricsNames {
+    pub driver_name: String,
```
rekby: `driver_name` is only initialized and overwritten; it is never read anywhere in the crate. Since this type is crate-private, the field triggers `dead_code`, which the CI `-D warnings` gate promotes to an error. Remove the field and its assignments unless it will actually be consumed.

ydb/src/client_metrics/names.rs
```diff
@@ -0,0 +1,72 @@
...
+            client_transaction_row_query_time_histogram: histogram!(description: "ydb transaction row query time histogram", "ydb_transaction_row_query_time_histogram", &labels),
+        }
+    }
+}
+
+impl MetricsNames {
+    pub fn new() -> Self {
+        let n: MetricsNames = Default::default();
+        n
+    }
```
rekby: The mandatory Clippy gate flags this as `clippy::let_and_return`. Return the default value directly.

```suggestion
    pub fn new() -> Self {
        Self::default()
    }
```

rekby (commented):

ydb/src/client_metrics/names.rs
rekby: Still present in the current head: `ClientBuilder::new()` immediately calls `MetricsNames::new()`, so every stored handle is bound before `build()`. Constructing the builder before installing a recorder still leaves the client with no-op handles; all new tests install the recorder first and miss this ordering.

ydb/src/client_metrics/names.rs
rekby: Still present: `MetricsNames::new()` registers all `driver_name="main"` handles before `with_driver_name()` replaces them. A custom-only driver therefore still creates the unwanted zero-valued `main` series, and there is no custom-driver assertion covering this.

ydb/src/client_builder.rs
rekby: Still present at this head: line 260 increments before the fallible interceptor/discovery setup and `Client::init(...).await`, so a failed build is counted as a newly created client.

ydb/src/client_query/mod.rs
rekby: Still present: the generated inherent methods only construct builders, while the increments remain exclusively in the `QueryExecutor` trait implementations. The new tests continue to use trait-qualified UFCS calls, so ordinary `qc.query_row(...)`, `tx.exec(...)`, and `tx.query_row(...)` remain uncounted.

ydb/src/client_query/mod.rs
rekby: Still present: `transaction_rollback` increments before checking for a transaction ID and lease, so an unmaterialized rollback is counted despite issuing no RPC. Conversely, `spawn_query_tx_rollback_on_drop` still performs a real rollback RPC without incrementing the counter.

ydb/src/client_metrics/names.rs
sergey-glyanets: Fixed

ydb/src/client_metrics/names.rs
sergey-glyanets: Reworked

ydb/src/client_metrics/names.rs
sergey-glyanets: Reworked.

ydb/src/client_query/builders.rs
sergey-glyanets: Replace trace with histogram metric.

ydb/src/client_query/builders.rs
```diff
@@ -211,10 +211,22 @@ impl<'a, T: FromYdbRow + 'a, S> IntoFuture for CallBuilder<'a, OneRow<T>, S> {
 
     fn into_future(mut self) -> Self::IntoFuture {
         Box::pin(async move {
+            let start = Instant::now();
```
rekby, line 214: Only the strict `OneRow` future is timed. Calling `query_row(...).optional()` changes the builder to `OptionalRow`, whose `IntoFuture` below has no histogram recording, so this common query-row path never contributes to either latency histogram. Move the timing into a shared query-row execution path or instrument both futures.

rekby (commented):

ydb/src/client_builder.rs
rekby: Creating the handles in `build` fixes the recorder/custom-label ordering, but `client_new_counter` still increments before the fallible interceptor/discovery setup and `Client::init(...).await`. A failed build is therefore still counted as a new client.

ydb/src/client_query/mod.rs
rekby: Still unresolved: the inherent methods generated in `builders.rs` only construct builders, while the increments remain solely in the `QueryExecutor` implementations. Ordinary `qc.query_row(...)`, `tx.exec(...)`, and `tx.query_row(...)` calls still bypass these counters; the tests continue to use trait-qualified UFCS calls.

ydb/src/client_query/mod.rs
rekby: Still unresolved: the counter increments before the transaction-ID/lease guard, so rollback without an actual RPC is counted. The real rollback RPC in `spawn_query_tx_rollback_on_drop` still has no increment.

ydb/tests/query_metrics.rs
```diff
@@ -0,0 +1,589 @@
...
+        .map(|mf| mf.metric.iter().map(|m| m.counter.value() as u64).sum())
+        .unwrap_or(0)
+}
+
+fn counter_label_value(metric_name: &str, label_key: &str) -> Option<String> {
+    let gathered = prometheus::default_registry().gather();
+    gathered
+        .iter()
+        .find(|mf| mf.name() == metric_name)
+        .and_then(|mf| mf.metric.first())
```
rekby, line 50: `METRICS_LOCK` only serializes these tests; it does not clear the process-global registry. Once a default-driver test has run, this family contains both `driver_name="main"` and `driver_name="custom"`, and `first()` is not guaranteed to select the custom series. The custom-label test therefore depends on test/collector order. Search `mf.metric` for the `driver_name="custom"` label (or use an isolated registry) instead.

rekby (commented):

ydb/src/client_builder.rs
rekby: Still unresolved: `client_new_counter.increment(1)` remains at the start of `build`, before fallible interceptor/discovery construction and `Client::init(...).await`, so failed builds are still counted.

ydb/src/client_query/mod.rs
rekby: Still unresolved: both macros in `builders.rs` continue to return `CallBuilder` directly, while counters are incremented only by the trait implementations. Normal `qc.query_row(...)`, `tx.exec(...)`, and `tx.query_row(...)` calls therefore still bypass the counters; the tests still use trait-qualified UFCS.

ydb/src/client_query/mod.rs
rekby: Still unresolved: `transaction_rollback` increments before its transaction-ID/lease guard, including when no rollback RPC occurs. Conversely, `spawn_query_tx_rollback_on_drop` can issue a rollback RPC without incrementing this counter.

ydb/tests/query_metrics.rs
rekby: Still unresolved: `counter_label_value` continues to inspect `mf.metric.first()`. The registry persists across these serialized tests, so a previously registered `driver_name="main"` series can be selected instead of `custom`, making this test order-dependent.

ydb/src/client_query/builders.rs
zeromem88: I suppose this is fixed because OptionalRow case covered below

rekby (commented):
OK

ydb/src/client_builder.rs
rekby: Still unresolved at the current head: `client_new_counter.increment(1)` runs at line 278 before fallible interceptor/discovery setup and `Client::init(...).await`, so failed builds are still exported as new clients.

ydb/src/client_query/mod.rs
rekby: Still unresolved: the inherent methods in `builders.rs` still construct `CallBuilder` directly, while the increments remain only in the `QueryExecutor` implementations. Ordinary `qc.query_row(...)`, `tx.exec(...)`, and `tx.query_row(...)` calls therefore still bypass these counters.

ydb/src/client_query/mod.rs
rekby: Still unresolved: `transaction_rollback` increments before checking for a usable transaction ID and lease, so no-RPC rollbacks are counted; `spawn_query_tx_rollback_on_drop` can issue the actual rollback RPC without incrementing the counter.

ydb/tests/query_metrics.rs
rekby: Still unresolved: `counter_label_value` continues to inspect `mf.metric.first()`. The global registry persists across these serialized tests, so the custom-driver assertion remains dependent on which labeled series is returned first.

rekby (approved):

PR #569, opened 2026-07-28
bench: add C++ Topic benchmark

aral01:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

This PR adds C++ topic benchmark, following existing Rust Topic benchmark.
C++ benchmark uses same scenarios and resulting schema.

### DIFF
--- a/benchmarks/sdk-compare/justfile
+++ b/benchmarks/sdk-compare/justfile
@@ -1,0 +1,74 @@
+set shell := ["bash", "-euo", "pipefail", "-c"]
+
+# Build the release benchmark and its local YDB C++ SDK.
+cpp-build:
+    @just _cpp-native-build native
+
+# Build and run the release benchmark. Defaults to local YDB.
+cpp-run scenario="scenarios/topic-smoke.json":
+    @just _cpp-native-build native >&2
+    @just _cpp-native-run native {{ quote(scenario) }}
+
+# Run native ASan and UBSan; defaults to the one-reader smoke scenario.
+cpp-asan scenario="scenarios/topic-smoke.json":
+    @just _cpp-native-build native-asan >&2
+    @ASAN_OPTIONS=abort_on_error=1:detect_leaks=0:symbolize=1 \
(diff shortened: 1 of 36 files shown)


benchmarks/sdk-compare/justfile
```diff
@@ -0,0 +1,74 @@
...
+        git clone https://github.com/ydb-platform/ydb-cpp-sdk.git "$sdk_source"
+        git -C "$sdk_source" checkout --detach 7e0839b71043ae4a6066ff934a62dd333d6cd265
+        git -C "$sdk_source" submodule update --init \
+            third_party/FastLZ \
+            third_party/aklomp-base64 \
+            third_party/api-common-protos \
+            third_party/jwt-cpp
+        git -C "$sdk_source" apply \
+            "$benchmark_root/cpp/patches/ydb-cpp-sdk.patch"
+    fi
```
rekby, line 45: The one-time SDK setup is guarded only by the existence of the checkout directory. `git clone` cleans up after itself on failure, but if any step *after* the clone fails — `checkout --detach`, `submodule update`, or `git apply` (e.g. a transient network error during submodule fetch, or the patch failing to apply after a future SDK bump) — the partially-prepared `$sdk_source` directory is left behind. Because `set -euo pipefail` aborts the recipe at that point but the directory now exists, the *next* `just cpp-build` sees `[[ ! -d "$sdk_source" ]]` as false, skips the whole `if` block, and builds against an un-checked-out / un-patched SDK, producing a confusing failure that only `rm -rf target/sdk-compare-ydb-cpp-sdk` resolves.

Consider making the setup atomic — e.g. prepare into a temp dir and `mv` into place only on full success, or drop a sentinel file (`.sdk-ready`) as the last step of the block and gate on that instead of the directory. That way an interrupted setup is retried on the next invocation rather than silently treated as complete.

rekby (commented):

aral01:
Moved to https://github.com/ydb-platform/ydb-sdk-benchmarks/pull/2

PR #579, opened 2026-08-03
feat: add `explain` method to QueryClient

DXist:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Can't verify a query and obtain its plan structure without running it if I use `QueryClient`.

Issue Number: N/A

## What is the new behavior?

`explain` allows adding query unit tests to validate the query text against the current database schema and check for plan issues like the presence of FullScan nodes.

## Other information

The original intent was to add all analysis exec modes (explain/validate/parse), so the low-level `raw_query_client` supports all three modes. Tests revealed that `parse` is not accepted by the latest local YDB and  `validate` is available only for scripts. Only `explain` is exposed to the QueryClient method.

The implementation was done using the [openspec](https://github.com/Fission-AI/OpenSpec/#see-it-in-action) workflow, and the produced spec artifacts are committed for reference. If desired, I can exclude them from the PR. Alternatively, after review, I can apply the `archive` command to persist their spec status from the transient one. The openspec commands/skills are not included as they could be easily installed via the `openspec` tool.


openspec/config.yaml
rekby, line 1: remove full folder with openspec please.

It is inconsistent, with codefor example: [initial_scpec.md](https://github.com/ydb-platform/ydb-rs-sdk/pull/579/changes#diff-741337ec71f1ccc96b989d99615d7fbeda551c5dd647f1a96114bc79ee8a8713R10) describe three methods, but [code](https://github.com/ydb-platform/ydb-rs-sdk/pull/579/changes#diff-0773b25fd9f1217bcea1a52d46c614a3f815fed9c359b149d48e559de1698970R28) doesn't contains them.

And I don't ready switch all development to specification driven develompent. Without the way spec will be inconsistent with code in the future event if it will be fixed now.

ydb/src/grpc_wrapper/raw_query_service/explain_validate_parse_query.rs
```diff
@@ -0,0 +1,269 @@
...
+
+/// Non-executing `Ydb.Query.ExecMode` values.
+///
+/// Only [`Explain`](Self::Explain) has a public entry point today. The Query Service rejects the
+/// other two on every server we can test against — `rpc_execute_query.cpp`'s `ParseQueryAction`
+/// has no `PARSE` branch, and `ValidateQuery` accepts only script query types, which the Query
+/// Service never uses. They are kept here so the decision can be made on PR review.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[allow(
+    dead_code,
```
rekby: No need dead_code in the code. Doesn't write constants, code, etc if it isn't need.

ydb/src/grpc_wrapper/raw_query_service/explain_validate_parse_query.rs
```diff
@@ -0,0 +1,269 @@
...
+        match self {
+            Self::Parse => ExecMode::Parse,
+            Self::Validate => ExecMode::Validate,
+            Self::Explain => ExecMode::Explain,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct RawExplainValidateParseRequest {
```
rekby: The PR should use `RawExecuteQueryRequest` and the `RawQueryServiceClient::execute_query` method.

The purpose is to simplify complex namespaces and module hierarchies, remove optional wrappers from fields, and make the types easier to use in production code. It is not intended to refactor the messages.

If any structures are missing fields required for the query, you can add them based on the original .proto files.

rekby (changes requested):
Thanks for the pr

It has some style comments before it can be merged.

ydb/src/grpc_wrapper/raw_query_service/explain_validate_parse_query.rs
DXist: @rekby Added `RawQueryStatsPlan` to `execute_query.rs` and removed the `explain_validate_parse_query.rs` module from the original PR.

openspec/config.yaml
DXist: Excluded `openspec` artifacts from the PR and rebased against the current master.

ydb/src/grpc_wrapper/raw_query_service/explain_validate_parse_query.rs
DXist: Removed and kept only `Execute` and `Explain` variants in the `RawExecMode` enum.

rekby (approved):

PR #584, opened 2026-08-03
docs: document all 25 examples in examples/README.md

sshaplygin:
## Problem

`ydb/examples/README.md` listed 8 of the 25 `cargo run --example` targets. Every topic, coordination, operation, auth and observability example was undiscoverable from the docs.

Separately, `ydb/examples/basic/README.md` instructed readers to run:

```bash
cargo run --example basic_query_series
```

No such target exists — the example is registered as `basic`, so the documented command fails.

## Change

- Group all 25 examples by the API they demonstrate, each with a one-line description and the environment variables it needs. The existing "Table Service API layers" table and the extra-dependency notes are kept as-is.
- Fix the wrong target name in `examples/basic/README.md`.
- Update the stale MSRV there (1.85 → 1.88).

## Verification

Checked mechanically that every example target reported by `cargo metadata` appears in the README — all 25 present.

### DIFF
--- a/ydb/examples/README.md
+++ b/ydb/examples/README.md
@@ -1,7 +1,15 @@
 # YDB SDK Examples
 
+Run any example with:
+
+```bash
+cargo run --example <name>
+```
+
 ## Common dependencies for examples
 All examples (but listed below) need local ydb started with docker-compose up with docker-compose.yaml from repository root dir.
+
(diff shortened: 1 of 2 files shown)


rekby (approved):

rekby:
Thanks for doc improvement

Issue #562, opened 2026-07-23
bug: panic in ClientBuilder


zeromem88:
# Bug Report

**Describe the bug**

**To Reproduce**

1. Go to: 
ydb/examples/auth-static-credentials.rs

2. Remove url scheme and run: 
```rust
let client = ClientBuilder::new_from_connection_string("localhost:2136?database=/Root/test".to_string()) // no  grpc://
    .unwrap()
    .with_credentials(StaticCredentials::new(
        "root".to_string(),
        "1234".to_string(),
        http::uri::Uri::from_static("grpc://localhost:2136/"),
        "/Root/test".to_string(),
    ))
    .client()?;
```
3. See panic:

```
create client...

thread 'main' (161602) panicked at ydb/src/client_builder.rs:380:24:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/std/src/panicking.rs:689:5
   1: core::panicking::panic_fmt
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/panicking.rs:80:14
   2: core::panicking::panic
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/panicking.rs:150:5
   3: core::option::unwrap_failed
             at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/option.rs:2236:5
   4: core::option::Option<T>::unwrap
             at /home/zerom/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:1016:21
   5: ydb::client_builder::ClientBuilder::parse_host_and_path
             at ./ydb/src/client_builder.rs:380:24
   6: ydb::client_builder::ClientBuilder::new_from_connection_string
             at ./ydb/src/client_builder.rs:246:24
   7: auth_static_credentials::main::{{closure}}
             at ./ydb/examples/auth-static-credentials.rs:10:9
   8: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
             at /home/zerom/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.38.1/src/runtime/park.rs:281:74
   9: tokio::runtime::coop::with_budget
[...]

PR #585, opened 2026-08-03
fix: remove panics from credentials production paths

sshaplygin:
## Problem

`credentials.rs` had five `unwrap`/`expect` calls on runtime-reachable paths, against the project rule that production code returns errors rather than panicking.

The most serious is in `CommandLineCredentials::create_token`:

```rust
result.status.code().unwrap()
```

`ExitStatus::code()` returns `None` when the child is terminated by a signal, so a `yc` helper killed by SIGKILL (OOM killer, timeout wrapper, operator) panics the caller's thread instead of reporting an auth failure.

This is separate from #562, which is the `ClientBuilder` panic.

## Change

| Site | Fix |
|---|---|
| `CommandLineCredentials::create_token` | Format the `ExitStatus` itself — it renders both exit codes and signals |
| `StaticCredentials::acquire_token` | Propagate the `get_auth_service` error with `?` |
| `ServiceAccountCredentials::build_jwt` | Return an error if the system clock is before the UNIX epoch |
| `GCEMetadata::new` / `MetadataUrlCredentials::new` | Build from the compile-time constant via a private `from_static_url`, so the infallible parse needs no `unwrap` |

No public signature changes.

## Test

Added `command_line_credentials_signal_killed_command`, which runs a self-killing shell script through `CommandLineCredentials`. Confirmed it reproduces the original failure — reverting only the `create_token` fix gives:

```
panicked at ydb/src/credentials.rs:268:38:
called `Option::unwrap()` on a `None` value
```

The test is `#[cfg(unix)]` and needs no YDB instance.

## Verification

```
cargo fmt --check && cargo clippy --workspace --all-targets --no-deps --exclude=ydb-grpc -- -D warnings
cargo test --workspace     # 215 passed, 88 ignored
```

### DIFF
--- a/ydb/src/auth_test.rs
+++ b/ydb/src/auth_test.rs
@@ -4,7 +4,10 @@ use tracing_test::traced_test;
 
 use crate::{
     YdbResult,
-    credentials::{CommandLineCredentials, StaticCredentials},
+    credentials::{
+        AnonymousCredentials, CommandLineCredentials, GCEMetadata, MetadataUrlCredentials,
+        ServiceAccountCredentials, StaticCredentials, unix_timestamp,
+    },
     pub_traits::Credentials,
     test_helpers::CONNECTION_STRING,
     test_integration_helper::create_password_client,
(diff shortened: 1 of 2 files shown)


sshaplygin:
Pushed `54c13ac` to raise patch coverage.

Codecov wants 80% on the diff; the changed lines started at **65.2%** because `credentials.rs` had no unit tests at all. Every changed line is now covered without a YDB instance:

| Changed code | Test |
|---|---|
| `CommandLineCredentials::create_token` error branch | signal-killed helper (already here), plus non-zero exit carrying stderr |
| `GCEMetadata::new` / `from_static_url` | asserts the default Google metadata endpoint |
| `MetadataUrlCredentials::new` / `default` | construction cannot panic |
| `StaticCredentials::acquire_token` `?` | closed loopback port returns a transport error instead of the old panic — refused immediately, ~8 ms, no network |

Also added the success path, `debug_string` not leaking the token, and rejection of an empty command line. The repeated temp-script setup moved into a small `ScriptCommand` helper that cleans up after itself.

**One production change beyond the original diff.** The JWT timestamp error branch was unreachable from a test — the `map_err` closure only runs if the clock is before the epoch — so it was written but never verified. Extracted it:

```rust
pub(crate) fn unix_timestamp(time: SystemTime) -> YdbResult<usize>
```

`build_jwt` now calls that, and both branches are tested directly. Happy to drop this if you would rather keep the diff purely at the call sites.

### Coverage

`credentials.rs`: lines 13.2% → 59.7%, regions 12.5% → 54.8%, functions 15.0% → 63.4%.

Patch coverage measures **85%** under `cargo llvm-cov --lib`. The three lines it still flags are inside `password_client_test`, which is `#[ignore]` — CI runs it with `--include-ignored` against a real YDB, so the reported number there should be higher.

`cargo fmt --check`, `clippy -D warnings` and `cargo test --workspace` all pass (227 passed, 88 ignored).

ydb/src/auth_test.rs
```diff
@@ -96,3 +102,240 @@ async fn password_client_test() -> YdbResult<()> {
...
+///
+/// `from_cmd` splits the command on whitespace, so the helper has to be a
+/// single path with no arguments - hence a script file rather than `sh -c`.
+#[cfg(unix)]
+struct ScriptCommand {
+    path: std::path::PathBuf,
+}
+
+#[cfg(unix)]
+impl ScriptCommand {
```
rekby: Зачем тут нужно генерировать внешние скрипты? Можно ли написать тесты внутри sdk?

sshaplygin: Убрал генерацию скриптов — тесты теперь целиком внутри SDK.

Вынес чистые части из `CommandLineCredentials`: `token_from_output` (готовый `Output` → токен) и `describe_token` (описание для логов). Тесты собирают `ExitStatus` сами: `from_raw(9)` — убийство сигналом, `from_raw(3 << 8)` — ненулевой код возврата, `ExitStatus::default()` — успешный путь. Ни временных файлов, ни дочерних процессов.

`#[cfg(unix)]` остался только на двух тестах с raw wait status (кодировка платформозависима), остальные стали кросс-платформенными.

Проверил, что регрессия ловится: с возвращённым `status.code().unwrap()` тест снова падает с паникой.

rekby (approved):

PR #586, opened 2026-08-03
docs: document core public API and guard it with deny(missing_docs)

sshaplygin:
## Problem

Building with `-W missing_docs` reports **497** undocumented public items, spread across most
modules. docs.rs therefore shows bare signatures for the types a user meets on their first query.

## Change

Document the core surface and make it non-regressable:

- **`errors.rs`** — `YdbOrCustomerError::to_ydb_error`, the five `YdbIssueSeverity` variants, the
  `YdbIssue` `issue_code`/`message` fields.
- **`result.rs`** — `ResultSet` with `rows`/`is_truncated`, `Row` with both `remove_field*`
  methods, `ResultSetRowsIter`.
- **`pub_traits.rs`** — `TokenInfo`, `Credentials` and both trait methods.
- **`waiter.rs`** — `Waiter` and `Waiter::wait`.

Then `#[deny(missing_docs)]` on those four module declarations in `lib.rs`, so the documented
surface cannot regress.

Where an item already carried implementation notes as plain `//` comments — `create_token`
(caching, background renewal, behaviour while renewal keeps failing) and `YdbIssueSeverity::Unknown`
(use for logs only, ask a maintainer to add the level) — those notes *become* the rustdoc rather
than being duplicated beside it. They were invisible on docs.rs before.

Per review, `issue_code` and the severity levels are documented as diagnostics rather than as
something to branch on: the code is more stable than the message but is not guaranteed per
scenario, and the SDK itself never reads `severity` — so neither replaces
`YdbStatusError::operation_status` in caller logic, and `Fatal` and `Error` are not presented as
if the SDK distinguished them.

That covers 21 of the 497 items. The other modules stay unguarded and can be converted one at a
time by the same pattern — the largest remaining are `client_topic/list_types.rs` (72 items),
`client_topic/client.rs` (46) and `table_requests.rs` (41).

## Verification

- `-W missing_docs` goes from 497 warnings to **476**, with **0** left in the four guarded
  modules.
- The guard bites: deleting one doc comment fails the build with
  `error: missing documentation for a method`.
- No new broken intra-doc links. `cargo doc -p ydb --no-deps` reports 24 warnings, all
  pre-existing and all in other modules (14 in `client_query/builders.rs`); none in the four
  touched files.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets --no-deps --exclude=ydb-grpc
  -- -D warnings`, `cargo test --workspace` (383 passed, 92 ignored) and `cargo test --doc -p ydb`
  (31 passed) all pass.

Rebased onto current `master`.

### DIFF
--- a/ydb/src/errors.rs
+++ b/ydb/src/errors.rs
@@ -32,6 +32,10 @@ impl YdbOrCustomerError {
         Self::Customer(Arc::new(Box::new(err)))
     }
 
+    /// Flatten into a [`YdbError`]
+    ///
+    /// A customer error is rendered through its [`Display`] impl and wrapped
+    /// into [`YdbError::Custom`].
     pub fn to_ydb_error(self) -> YdbError {
         match self {
             Self::YDB(err) => err,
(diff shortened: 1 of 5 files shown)


ydb/src/errors.rs
```diff
@@ -229,12 +233,22 @@ impl YdbStatusError {
 #[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
 #[cfg_attr(not(feature = "force-exhaustive-all"), non_exhaustive)]
 pub enum YdbIssueSeverity {
+    /// The operation cannot continue.
     #[default]
     Fatal,
+
+    /// The operation failed.
```
rekby: Тут непонятно различие между этими двумя ошибками. И в fatal и в error - операция дальше не продолжается.

ydb/src/errors.rs
```diff
@@ -272,7 +286,12 @@ impl From<u32> for YdbIssueSeverity {
 #[cfg_attr(not(feature = "force-exhaustive-all"), non_exhaustive)]
 // Combine with YdbStatusError?
 pub struct YdbIssue {
+    /// Server-defined issue code
+    ///
+    /// More stable than the message - prefer it for programmatic checks.
```
rekby: На эти коды тоже нельзя полагаться в общем случаи. Они более стабильны, чем текст. Но гарантии их стабильности и неизменности на конкретных сценариях нет. Это низкоуровневые штуки, больше для диагностики. И общая рекомендация - не завязываться на них.

rekby (changes requested):

PR #587, opened 2026-08-03
dev: replace unmaintained derivative crate with manual impls

sshaplygin:
## Problem

`derivative` is flagged unmaintained by [RUSTSEC-2024-0388](https://rustsec.org/advisories/RUSTSEC-2024-0388) (upstream is no longer maintained; the advisory suggests `derive_more`, `derive-where` or `educe`). Every `cargo audit` run against a project depending on `ydb` reports it.

## Change

It was used for one thing: derive `Debug` while skipping fields that are not `Debug` themselves. Each use becomes a hand-written impl listing the remaining fields, which produces identical output, and the dependency is dropped.

| Type | Skipped fields |
|---|---|
| `RacyRoundRobinState` | both `ConnectionTask` queues |
| `GrpcConnectionManagerGeneric` | `interceptor` |

`GrpcConnectionManagerGeneric` also used `derivative(Clone(bound = "BalancerT: Clone"))`, so its `Clone` becomes a manual impl too, preserving the balancer-only bound — `ConnectionT` sits behind an `Arc` and does not need `Clone`. Its `Debug` keeps the `ConnectionT: Debug` bound that `derivative` generated implicitly.

No behaviour change: `derivative`'s `Debug = "ignore"` omits the field entirely, which is what `debug_struct(..).field(..).finish()` emits.

A third use, on `TableSession`, was in the branch when it was opened. `master` has since removed that struct's non-`Debug` fields in the query/table session refactor (#621), so after the rebase there is nothing left to convert there.

## Tests

`derivative(Debug = "ignore")` enforced the skipped fields structurally; a hand-written `debug_struct` chain can silently start or stop reporting a field. Nothing in the suite formatted either type before, so each now has a test pinning both halves of the contract: the fields that must appear, and the non-`Debug` fields that must not. Two further tests pin the `BalancerT: Clone`-only bound and that clones keep sharing one pool.

## Verification

```
cargo tree -p ydb -e normal -i derivative
#   error: package ID specification `derivative` did not match any packages
cargo fmt --check
cargo test --workspace     # 385 passed, 0 failed, 92 ignored
```

Integration tests (`#[ignore]`, need `YDB_CONNECTION_STRING`) were not run.

The `Cargo.lock` diff is exactly the removal of `derivative 2.2.0` — no incidental version bumps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -493,17 +493,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "derivative"
-version = "2.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fcc3dd5e9e9c0b295d6e1e4d811fb6f157d5ffd784b8d202fc62eac8035a770b"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 1.0.109",
-]
-
-[[package]]
 name = "derive_builder"
 version = "0.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3937,7 +3926,6 @@ dependencies = [
  "bon",
  "chrono",
  "decimal-rs",
- "derivative",
  "derive_builder",
  "flate2",
  "futures-util",
--- a/ydb/Cargo.toml
+++ b/ydb/Cargo.toml
@@ -29,7 +29,6 @@ chrono = { version = "0.4", default-features = false, features = [
     "serde",
 ] }
 decimal-rs = { version = "0.1", features = ["serde"] }
-derivative = "2"
 bon = "3"
 derive_builder = "0.12.0"
 futures-util = "0.3"
--- a/ydb/src/connection_pool.rs
+++ b/ydb/src/connection_pool.rs
@@ -1,11 +1,11 @@
 use crate::{GrpcOptions, YdbError, YdbResult};
-use derivative::Derivative;
 use futures_util::FutureExt;
 use futures_util::stream::FuturesUnordered;
 use http::Uri;
 use http::uri::{Authority, Scheme};
 use itertools::Either;
 use std::collections::{HashMap, HashSet, VecDeque};
+use std::fmt::{Debug, Formatter};
 use std::net::IpAddr;
 use std::sync::Arc;
 use std::task::Poll;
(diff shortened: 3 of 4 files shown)


sshaplygin:
_Updated after the rebase onto `master` (`fe2d450`); the branch is now `935e50e` + `a9a8f81`._

Pushed `a9a8f81` to cover the hand-written impls. Nothing in the suite ever formatted the types whose derives this PR replaces, so the impls shipped untested.

That is worth closing for its own sake, not just for the coverage number — `derivative(Debug = "ignore")` enforced the skipped fields structurally, while a hand-written `debug_struct` chain can silently start or stop reporting a field. Each test pins both halves of the contract.

| Type | Must report | Must never report |
|---|---|---|
| `RacyRoundRobinState` | addrs, first_connection | the two connection queues |
| `GrpcConnectionManagerGeneric` | balancer, connections_pool, database | `interceptor` |

The `TableSession` row that this comment listed before is gone: the rebase picked up #621, which reworked `TableSession` onto `SessionPoolLease` and removed the non-`Debug` fields (`channel_pool`, `on_drop_callbacks`) that the `derivative` derive existed to skip. There is nothing left to convert there, so both the impl and its test dropped out of the branch.

Two extra tests pin the `Clone` side: the `BalancerT: Clone`-only bound that `derivative(Clone(bound = ...))` gave us, and that clones keep sharing one `Arc` pool — checked for both `NoBalancer` and the `SharedLoadBalancer` used in production.

Nothing contacts a server: `RacyRoundRobinState` uses a lazily connected channel (async only because `connect_lazy` needs a reactor).

The tests are not vacuous — dropping one `.field(..)` from each impl fails the matching test:

```
first_connection missing: RacyRoundRobinState { addrs: {127.0.0.1} }
database missing: GrpcConnectionManagerGeneric { balancer: NoBalancer, connections_pool: ... }
```

### Coverage

Measured locally with `cargo llvm-cov --lib` on the rebased branch, comparing `935e50e` (impls only) with `a9a8f81` (impls + tests) — not Codecov's own run:

- patch coverage of the branch's added lines: **39.1% (9/23) → 100% (86/86)**
- `grpc_connection_manager.rs` 69.8% → 89.4% lines, `connection_pool.rs` 37.9% → 46.4%

`session.rs` is untouched by the branch now and stays at 39.5%.

`cargo fmt --check`, `cargo clippy --workspace --all-targets --no-deps --exclude=ydb-grpc -- -D warnings` and `cargo test --workspace` all pass (385 passed, 0 failed, 92 ignored). The ignored ones are the integration tests that need `YDB_CONNECTION_STRING`; they were not run.

PR #588, opened 2026-08-03
build: declare dependencies in the workspace manifest and enforce the graph in CI

sshaplygin:
Reworked. The first version of this PR bumped `reqwest` to collapse the second HTTP stack by hand. That fixes one symptom; this version fixes the reason it happened and makes CI keep the dependency graph in shape, so the next duplicate does not have to be spotted by a human.

It also folds in the dependency upgrades from #625 (@rinatshigapov), which arrived at the same manifests from the other side — see **Relationship to #625** below for what was taken, what was not, and why.

## Problem

Version requirements were spread over five manifests (`ydb`, `ydb-grpc`, and the three SLO crates), and only the protobuf/gRPC family was declared in `[workspace.dependencies]`. Nothing checked that two members ask for the same crate with the same requirement, and nothing checked the resolved tree at all. Three consequences, all present on `master`:

**1. A pin in a test workload decided the version for the published crate.** `tests/slo/*` pinned `tokio = "=1.38.1"` and `tokio-util = "=0.7.11"`. Since the workspace resolves once, `ydb` — which declares `tokio = "1.22"` — was built and tested against tokio 1.38.1, including the leg of `rust-tests.yml` that is supposed to cover the current stack. tokio 1.38.1 is affected by RUSTSEC-2025-0023.

**2. Two parallel stacks of the same libraries.** `reqwest 0.11` is built on `hyper 0.14` / `http 0.2` / `rustls 0.21`, next to tonic's `hyper 1` / `http 1` / `rustls 0.23`. 21 crates were present in more than one version in `cargo tree -p ydb -e normal`; downstream users compile, link and audit all of them.

**3. A `Cargo.lock` nobody refreshes.** `cargo deny check advisories` against the committed lock reports 12 advisories on nine crates, plus the `derivative` entry this PR ignores explicitly:

| Crate | Advisory | |
|---|---|---|
| `anyhow` 1.0.100 | RUSTSEC-2026-0190 | unsound `Error::downcast_mut()` |
| `bytes` 1.10.1 | RUSTSEC-2026-0007 | integer overflow in `BytesMut::reserve` |
| `crossbeam-epoch` 0.9.18 | RUSTSEC-2026-0204 | invalid pointer dereference |
| `h2` 0.3.27, 0.4.12 | RUSTSEC-2026-0258 | unbounded empty DATA frames |
| `rand` 0.8.5 | RUSTSEC-2026-0097 | unsound with a custom logger |
| `rustls-pemfile` 1.0.4 | RUSTSEC-2025-0134 | unmaintained (reached only through `reqwest` 0.11) |
| `rustls-webpki` 0.101.7, 0.103.7 | RUSTSEC-2026-0049, -0098, -0099, -0104 | CRL and name-constraint handling |
| `time` 0.3.44 | RUSTSEC-2026-0009 | stack exhaustion parsing RFC 2822 |
| `tokio` 1.38.1 | RUSTSEC-2025-0023 | broadcast channel clones in parallel without `Sync` |
| `derivative` 2.2.0 | RUSTSEC-2024-0388 | unmaintained |

All of them except `derivative` are fixed by requirements that were already declared — they only needed a resolution nobody triggered.

### 1. One requirement per crate, in the root manifest

`[workspace.dependencies]` now holds all 59 third-party dependencies of all six members, grouped by role and commented where a version is not free to move. Members inherit with `{ workspace = true }` and may only add `features`:

```toml
# ydb/Cargo.toml
tokio = { workspace = true }
# ydb-grpc/Cargo.toml
tonic = { workspace = true, features = ["tls-native-roots", "tls-webpki-roots"] }
```

The exact pins (`=1.38.1`, `=0.7.11`, `=4.2.7`) are gone; the SLO crates use the same requirements as everything else. Path dependencies (`ydb`, `ydb-grpc`, `slo-framework`) are declared in the workspace table too, so `ydb-grpc`'s version lives in one place at publish time.

### 2. Requirements raised where they were the cause of a duplicate

| Crate | Was | Now | Collapses |
|---|---|---|---|
| `reqwest` | 0.11 | 0.13 | `hyper`, `http`, `http-body`, `h2`, `rustls`, `rustls-webpki`, `tokio-rustls`, `webpki-roots`, `sync_wrapper`, `base64`, `bitflags`, `core-foundation`, `rustls-pemfile` |
| `jsonwebtoken` | 9.3 | 11 | drops `ydb`'s direct `ring` dependency |
| `rand` | 0.8 | 0.9 | aligns with the OpenTelemetry stack |
| `itertools` | 0.10 | 0.14 | `itertools` (0.14 is what `prost-derive` uses) |
| `tower` | 0.4 | 0.5 | `tower` (0.5 is what `tonic` and `reqwest` use) |
| `strum` | 0.21 | 0.28 | one of the two `syn` 1 users |
| `derive_builder` | 0.12 | 0.20 | the other `syn` 1 user, and `strsim` 0.10 |
| `http` | 1.1 | 1.5 | — |
| `tokio` | =1.38.1 / 1.22 | 1.38 | `socket2` |
| `clap` | =4.2.7 | 4.6 | `bitflags` 1 in the SLO workload |
| OpenTelemetry (SLO) | 0.27 | 0.32 | `prost` 0.13, `tonic` 0.12, `thiserror` 1, the second `opentelemetry` stack |

`reqwest` 0.13 renamed `rustls-tls` and made the new `rustls` feature pull `aws-lc-rs` and `rustls-platform-verifier`. It is taken with `rustls-no-provider` instead, so the crypto provider keeps coming from `tonic`'s `tls-ring` and `ring` stays the only provider in the tree — `rustls` 0.23 fails at request time, not at build time, when more than one is linked without an explicit choice. There is no `aws-lc-rs` anywhere in `Cargo.lock`.

Moving the SLO workload to OpenTelemetry 0.32 needed the exporter to be handed a preconfigured HTTP client (`rustls::ClientConfig` with the platform verifier), because `opentelemetry-otlp`'s own `reqwest-rustls` feature maps onto the `reqwest` feature that brings `aws-lc-rs` in. `Resource::new` and the runtime-bound `PeriodicReader` are gone in 0.32, so `metrics.rs` uses `Resource::builder()` and the thread-based reader.

`Cargo.lock` is refreshed to current compatible versions. No requirement in any manifest was widened to make that possible.

### 3. `deny.toml` — the policy as an enforced rule

- `bans.multiple-versions = "deny"`, `wildcards = "deny"` (with `allow-wildcard-paths` for the `publish = false` members that depend on `ydb` by path).
- `[bans].skip` lists every duplicate the workspace cannot collapse, each with the crates that keep the old version alive. Nothing is skipped without a reason next to it.
- `skip-tree` for the Windows-only `windows-sys` family.
- `[licenses].allow` is the set of permissive licenses actually present in the published graph. `[sources]` restricts the tree to crates.io.
- `[advisories].ignore` has exactly one entry, RUSTSEC-2024-0388, with the reason and the way out.

### 4. `.github/workflows/dependencies.yml`

Two invocations, on push, on pull request and on a nightly cron:

```bash
# the graph a user of the published crates compiles
cargo deny --locked --all-features --exclude-unpublished --exclude-dev \
    check bans licenses sources -D unmatched-skip -D unnecessary-skip
# advisories over everything built in this repository
cargo deny --locked --all-features check advisories
```

`--exclude-unpublished` drops the `publish = false` SLO binaries, `--exclude-dev` drops dev-dependencies: the single-version rule applies to what users actually compile, not to the test harness. Advisories cover everything.

`-D unmatched-skip -D unnecessary-skip` turns a `[bans].skip` entry into a build failure once it stops matching, or once the crate it covers is back to a single version, so an entry has to be deleted rather than left behind. The skip list cannot silently become a blanket allowance. It already earned its keep: `prettyplease` left the list because of it.

The cron matters on its own: an advisory is published without anyone touching the repository, and `Cargo.lock` is committed, so nothing else would ever notice.

### 5. `.github/scripts/check_workspace_deps.py`

`cargo` accepts a version requirement in a member manifest, so the rule above has to be checked. The script fails when a member declares its own requirement, and when a `[workspace.dependencies]` entry is not inherited by any member (a leftover entry is a version that looks maintained but is dead). It reads the manifests with `tomllib`, no extra tooling.

The exemption table in the script is empty: with the SLO workload moved to OpenTelemetry 0.32, every dependency of every member is inherited. An entry added there later has to carry its reason.

### 6. Documentation

`.agents/context/techContext.md` gets the policy, the commands, the new workflow row, and the rule for holding a version back — in `Cargo.lock` with a comment, never as an upper bound in `[workspace.dependencies]`, because a bound there is published in the `ydb` manifest. `AGENTS.md` gets the one-line rule and the extra gate for changes that touch `Cargo.toml`, `Cargo.lock` or `deny.toml`.

## Effect

`cargo tree -p ydb -e normal` — the graph a user of the crate builds:

- crates present in more than one version: **21 → 2**
- total crates: **222 → 201**

The same two are all that is left across the whole workspace, dev-dependencies and SLO binaries included (**29 → 2**): with the workload on the same OpenTelemetry as the examples, the test harness no longer drags a second `prost`, `tonic` and `thiserror` behind it.

`cargo deny --all-features check advisories` over the whole workspace: **12 advisories → 0**, with the single documented ignore for `derivative`.

What is still duplicated, as `cargo deny` sees it (its graph also covers build-dependencies and every target, so it lists two entries that `cargo tree` on Linux or macOS does not). All of it is in `deny.toml` with the same wording:

| Crate | Why |
|---|---|
| `getrandom` | Three versions, one per consumer: `ring` (through `rustls`) is on 0.2, `rand` 0.9 on 0.3, `uuid` 1 on 0.4. None of them is ours to move. |
| `r-efi` | UEFI-only dependency of `getrandom`, one version per `getrandom` major. |
| `syn` | The proc-macro ecosystem is between `syn` 2 and `syn` 3; `derivative` is still on `syn` 1. Build-time only. |
| `hashbrown` | Build-time only: `indexmap` is on 0.17, `petgraph` (through `prost-build`) on 0.15. |

`darling` is no longer among them, see below.

## Relationship to #625

#625 (@rinatshigapov) upgrades the same dependencies from the other side: its author hit `rustls-webpki` 0.101.7 in an application that depends on `ydb`, which has no `anchor_from_trusted_cert` and made TLS trust anchors impossible to configure. Both PRs rewrite the same manifests, so rather than leaving them to conflict, its changes are folded in here — `reqwest` 0.13, `jsonwebtoken` 11, `rand` 0.9 with its call sites, `strum` 0.28, `http` 1.5, `clap` 4.6, and the OpenTelemetry 0.32 migration of the SLO workload including the preconfigured-TLS-backend exporter.

Four deliberate differences:

**`tonic` keeps `tls-ring`.** #625 removes it so the application picks the crypto provider. That makes `ydb` panic at the first TLS connection for every user who does not add `rustls` with a provider feature to their own manifest — a breaking change for the published crate and a call for the maintainers, not something to slip in with a dependency sweep. The `reqwest` half of it is kept: `rustls-no-provider` instead of 0.13's `rustls`, which would add `aws-lc-rs` next to the `ring` that `tonic` already selects. There is no `aws-lc-rs` anywhere in `Cargo.lock`, and `ydb` no longer depends on `ring` directly either — `jsonwebtoken` 11 dropped it.

**No `[patch.crates-io]`.** #625 patches four OpenTelemetry crates to a git revision. The released 0.32 turned out to be enough once the exporter is handed a preconfigured HTTP client, which avoids both a git source in the tree and an exception in `[sources]`.

**`bon` is held at 3.9.x in `Cargo.lock` instead of being constrained to "3.5".** The requirement in #625 does not have the effect it describes: a caret requirement still resolves to the newest 3.x, and `bon` 3.10 moved to `darling` 0.24 while `derive_builder` 0.20 — its last release — is capped at `darling` 0.20. `bon` 3.8.2 through 3.9.3 accept `>=0.20.0, <0.24.0`, so holding the lock there puts both builder crates on `darling` 0.20.11 and removes the duplicate. The bound is deliberately not published in the `ydb` manifest, where it would constrain applications that use `bon` themselves; the reason sits in a comment next to the requirement, and a `cargo update` that raises `bon` fails the bans check rather than silently reintroducing the duplicate.

**The `=` pins on `tokio`, `tokio-util` and `clap` are removed** rather than kept.

## Notes for review

- **Public API.** `YdbOrCustomerError` has `From` impls for `reqwest::Error` and `strum::ParseError` (`ydb/src/errors.rs`). After this change those impls refer to `reqwest` 0.13 and `strum` 0.28 types, so a caller who relies on `?` converting their own `reqwest::Error` has to be on 0.13 as well. Same for the traits `strum` and `derive_builder` derive on public types.
- **TLS is unchanged for users.** `ring` is still the only crypto provider in the published graph, selected by `tonic`'s `tls-ring` as before. What changes is that `reqwest` no longer brings its own: it is on `rustls-no-provider`.
- **Declared minimum for `tokio` moves from 1.22 to 1.38.** 1.22 was never built or tested — the SLO pin made 1.38.1 the resolved version for years. The new floor is the one that has actual coverage.
- **A latent bug fell out of the `reqwest` bump.** `client_coordination::session::controller` imported `tracing::log::trace`, which only resolved because `reqwest` 0.11 happened to enable the `log` feature of `tracing`. It now uses `tracing::trace`. Nothing else in the crate depended on a feature it did not declare.
- The branch name is left over from the first version of this PR.

## Verification

```
cargo fmt --all --check
taplo fmt                       # no changes
cargo test --workspace          # 363 passed, 92 ignored (integration)
cargo +1.88 check --workspace --all-targets --locked
cargo +1.88 check -p ydb --lib --locked --all-features
python3 .github/scripts/check_workspace_deps.py
cargo deny --locked --all-features --exclude-unpublished --exclude-dev check bans licenses sources -D unmatched-skip -D unnecessary-skip
cargo deny --locked --all-features check advisories
```

All pass. The `-D unmatched-skip` gate was checked by adding a skip entry for a crate that is not in the tree: the run fails, as intended. `-D unnecessary-skip` fired on its own during the work and removed the `prettyplease` entry. Integration tests were not run locally — they need a YDB instance; CI covers them. The SLO workloads were compiled but not executed; the `SLO` label is the way to exercise them, and this PR changes their tokio, clap and OpenTelemetry versions.


DXist:
+1

Old reqwest downgrades rustls-webpki to v0.101.7, which doesn't have `webpki::anchor_from_trusted_cert`, needed for TLS trust anchor configuration.

It's also possible to bump it to "0.13" if opentelemetry is installed from the current main branch. Done it [here](https://github.com/ydb-platform/ydb-rs-sdk/pull/625/changes#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R47).

DXist:
>  That makes ydb panic at the first TLS connection for every user who does not add rustls with a provider feature to their own manifest — a breaking change for the published crate and a call for the maintainers, not something to slip in with a dependency sweep.

@sshaplygin , is it possible to introduce `rustls-ring` feature and include it in the default features to maintain backward compatibility?

Users wishing to change the crypto provider could opt-out from the default features. For convenience and symmetry, it's possible to add the `rustls-aws-lc-rs` feature as well.

PR #590, opened 2026-08-03
fix: wake coordination waiters when the session stream fails

sshaplygin:
> **Stacked on the request-registration fix.** This branch contains that commit as well, so the diff below shows both. Please merge that one first; this PR then reduces to its own commit. I could not set it as the base because it does not exist in this repository yet.

## Problem

When a coordination session's stream failed, the loop cancelled its token and returned — but nothing told the callers already waiting on a response. Their entries stayed in `active_requests`, holding a sender that would never be used, so `rx.recv()` never resolved. A caller blocked in `acquire_semaphore` during a disconnect waits forever.

`acquire_semaphore` also did:

```rust
let response = rx.recv().await.unwrap();
```

So in the one case where the channel *did* close, the caller panicked instead of getting an error.

## Change

- `RequestController::close` marks the controller closed and clears `active_requests`. Dropping the senders resolves every waiting `recv()` with `None`, so blocked callers wake up.
- `send` refuses new requests on a closed controller. The `closed` flag is checked once up front and again under the `active_requests` lock, so a `close` racing an in-flight `send` cannot leave a waiter behind after the map was cleared.
- `MethodControllers::close` fans this out across all six per-method controllers, and the session loop calls it on stream failure before cancelling its token.
- `acquire_semaphore` maps the closed channel to an error instead of unwrapping it.

## Test

`closing_controller_unblocks_in_flight_request` registers a request, closes the controller, and asserts the caller's `recv()` completes with `None` inside 50 ms rather than hanging, and that subsequent sends are rejected.

## Verification

```
cargo fmt --check
cargo test --workspace     # 216 passed, 88 ignored
```

### DIFF
--- a/ydb/src/client_coordination/session/controller.rs
+++ b/ydb/src/client_coordination/session/controller.rs
@@ -16,6 +16,7 @@ pub trait IdentifiedMessage {
 
 pub struct RequestController<Response: IdentifiedMessage> {
     last_req_id: atomic::AtomicU64,
+    closed: atomic::AtomicBool,
     messages_sender: mpsc::UnboundedSender<SessionRequest>,
     active_requests: Arc<Mutex<HashMap<u64, tokio::sync::mpsc::UnboundedSender<Response>>>>,
 }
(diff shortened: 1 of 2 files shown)


sshaplygin:
Depends on #589 — this branch is stacked on it and its diff includes that commit. Merging #589 first reduces this PR to its own change.

PR #592, opened 2026-08-03
fix: coalesce forced rediscovery instead of spawning a task per failure

sshaplygin:
## Problem

`TimerDiscovery::pessimize` spawns a forced rediscovery whenever at least half the nodes are pessimized:

```rust
if pessimized_nodes_count > 0 && pessimized_nodes_count >= state.original_nodes.len() / 2 {
    tokio::spawn(async move { ... state.discovery_now().await ... });
}
```

`pessimize` is called per failing request. During an outage — exactly when that condition holds — every failure spawns another task. They all queue on the same `discovery_lock`, so the work is serialised but the tasks are not: the SDK piles up rediscovery tasks against a cluster that is already struggling, which is the opposite of what a backoff should do.

## Change

Guard the spawn with an `AtomicBool` on `DiscoverySharedState`. The first caller claims the refresh via `compare_exchange`; further callers observe one already in flight and skip. The flag is released once `discovery_now` returns, so the next burst can trigger a fresh refresh.

This coalesces the storm without changing when a refresh happens, or the existing `discovery_lock` serialisation.

## Test

`forced_discovery_coalesces_requests_until_completion` asserts a second claim is refused while one is in flight, and that a claim succeeds again after completion.

## Note for reviewers

The flag is cleared after `discovery_now().await` returns, including on `Err`. If that spawned task were cancelled — for example at runtime shutdown — the flag would stay set and forced rediscovery would be disabled for the remaining life of the state. That seemed acceptable given the task only runs for as long as the driver does, but say the word if you would rather see it held by a guard that clears on drop.

## Verification

```
cargo fmt --check
cargo test --workspace     # 215 passed, 88 ignored
```

### DIFF
--- a/ydb/src/discovery.rs
+++ b/ydb/src/discovery.rs
@@ -513,13 +513,20 @@ mod test {
 
     use crate::GrpcOptions;
     use crate::client_common::{DBCredentials, TokenCache};
-    use crate::discovery::{Discovery, DiscoverySharedState, DiscoveryState, NodeInfo};
-    use crate::errors::YdbResult;
+    use crate::discovery::{
+        Discovery, DiscoverySharedState, DiscoveryState, NodeInfo, StaticDiscovery, TimerDiscovery,
+    };
+    use crate::errors::{YdbError, YdbResult};
     use crate::grpc_connection_manager::{DiscoveryConnectionManager, NoBalancer};
     use crate::grpc_wrapper::auth::AuthGrpcInterceptor;
+    use crate::grpc_wrapper::raw_discovery_client::EndpointInfo;
+    use crate::grpc_wrapper::raw_services::Service;
     use crate::grpc_wrapper::runtime_interceptors::MultiInterceptor;
     use crate::test_helpers::test_client_builder;
+    use crate::waiter::Waiter;
+    use futures_util::StreamExt;
     use std::sync::Arc;
+    use std::sync::atomic::Ordering;
     use std::time::{Duration, Instant};
 
     fn discovery_shared_state() -> YdbResult<DiscoverySharedState> {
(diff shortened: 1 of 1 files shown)


sshaplygin:
Pushed `54d72f3` raising `ydb/src/discovery.rs` past the 80% target.

| Metric | Before | After |
|---|---|---|
| Lines | 34.4% | **87.3%** |
| Regions | 38.8% | **88.8%** |
| Functions | 31.6% | **94.9%** |

Those are whole-file numbers, so they include the inline test module. Excluding it, the production code alone goes **36.3% → 97.4%** of lines.

The module was almost entirely reachable only through the two `#[ignore]` tests that need a live YDB. Now covered offline:

- **`DiscoveryState`** — node exposure, empty default, pessimization and its idempotence, the fallback that restores every node once all are pessimized, `with_node_info`.
- **Endpoint conversion** — http/https by the ssl flag, and rejection of a malformed authority.
- **`StaticDiscovery`** — state, waiter, no-op pessimization, empty subscription, malformed endpoint.
- **`DiscoverySharedState`** — no state before first discovery, `state()` resolving on publish, `wait()` surfacing a stored error, `subscribe()` filtering failures out, pessimization updating the published state.
- **`background_discovery`** — exits once the shared state is dropped.

### On the coalescing itself

The first version of the storm test spawned the refresh but never waited for it, so lines 252–256 — the spawned task body, the core of this PR — stayed uncovered even though the test passed. It now waits for `forced_discovery_in_flight` to clear, which only happens after the task has run `discovery_now` and called `complete_forced_discovery`. It then asserts a second storm can schedule a fresh refresh, which is the property the flag exists to provide.

Nothing contacts a server: refreshes point at a closed loopback port, refused immediately.

The one production branch still uncovered is the token-waiter error path in `background_discovery` (lines 392–394), which needs a failing token source to reach.

`cargo fmt --check`, `clippy -D warnings` and `cargo test --workspace` pass (234 passed, 88 ignored).

PR #619, opened 2026-08-13
refactor: query stream session ownership

aral01:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

- OpenedQueryStream has session lease owning ambiguity. Some -> owns session, None -> Query is session-less or session belongs to transaction.
- TxState carries to resources, so those resources are stores as Optional values in transaction context (session/transaction_id)

Issue Number: N/A

## What is the new behavior?

- QueryStream manages its session owning in explicit enum with descriptive naming. 
- TransactionState owns the resources. Active state owns lease and ServerState.
- Query completion, failure and drop explicitly return or discard the owned session.
- Failure of the first dispatched query without a confirmed transaction ID makes the transaction ambiguous and discards its session, preventing a false commit.

### DIFF
--- a/ydb/src/client_query/exec.rs
+++ b/ydb/src/client_query/exec.rs
@@ -4,7 +4,7 @@ use std::time::{Duration, Instant};
 
 use tokio::time::timeout;
 
-use crate::errors::{Idempotency, TxErrorOutcome, YdbError, YdbResult};
+use crate::errors::{Idempotency, YdbError, YdbResult};
 use crate::grpc_connection_manager::GrpcConnectionManager;
 use crate::grpc_wrapper::raw_query_service::client::RawQueryClient;
 use crate::grpc_wrapper::raw_query_service::execute_query::RawExecuteQueryRequest;
@@ -77,9 +77,9 @@ pub(crate) enum TxState {
     Committed,
     /// Rollback path was chosen and the SDK must not report a commit.
     RolledBack,
-    /// A definitive operation failure ended the local transaction attempt.
+    /// A transaction operation failed; retry policy decides whether to start a new attempt.
     AttemptFailed(YdbError),
-    /// A commit or rollback RPC returned an error, so the local end attempt was not confirmed.
+    /// A commit or rollback RPC failed after dispatch, so its outcome was not confirmed.
     Undetermined(YdbError),
 }
 
@@ -164,18 +164,15 @@ impl TxExecContext {
         }
     }
 
-    fn finish_dispatched_error(&mut self, error: &YdbError) -> YdbResult<()> {
-        let outcome = error.tx_error_outcome();
-        let replacement = match outcome {
-            TxErrorOutcome::Ended | TxErrorOutcome::MayRemainActive => {
-                TxState::AttemptFailed(error.clone())
-            }
-            TxErrorOutcome::Undetermined => TxState::Undetermined(error.clone()),
-        };
-        let active = self.replace_active(replacement)?;
-        if outcome == TxErrorOutcome::Ended && !error.requires_session_discard() {
-            active.lease.return_to_pool();
+    fn fail_attempt(&mut self, error: &YdbError) -> YdbResult<()> {
+        let active = self.replace_active(TxState::AttemptFailed(error.clone()))?;
+
+        if error.requires_session_discard() {
+            // Dropping a lease without returning it schedules session cleanup.
+            return Ok(());
         }
+
+        release_unfinished_tx(self.connection_manager.clone(), active);
         Ok(())
     }
 }
@@ -529,7 +526,7 @@ pub(crate) async fn tx_ensure_begin(tx: &mut TxExecContext) -> YdbResult<()> {
             Ok(())
         }
         Err(err) => {
-            tx.finish_dispatched_error(&err)?;
+            tx.fail_attempt(&err)?;
             Err(err)
         }
     }
@@ -569,10 +566,10 @@ async fn tx_before_commit(tx: &mut TxExecContext) -> YdbResult<()> {
     Ok(())
 }
 
-/// Apply a query error to the retained session and transaction state.
+/// Finish a failed query attempt and release its transaction resources.
 pub(crate) fn tx_handle_query_error(tx: &mut TxExecContext, err: &YdbError) {
     if tx.state.is_active()
-        && let Err(error) = tx.finish_dispatched_error(err)
+        && let Err(error) = tx.fail_attempt(err)
     {
         tracing::error!(%error, "failed to finish transaction after query error");
     }
@@ -597,6 +594,7 @@ pub(crate) async fn tx_begin_stream(
         "implicit_session is only available on QueryClient one-shot builders"
     );
     tx.active()?;
+    let commit_at_end = opts.commit_tx.unwrap_or(false);
     let effective_timeout = resolve_effective_timeout(tx.retry_deadline, opts.timeout);
     let mut query_dispatched = false;
     let result: YdbResult<ExecuteQueryStream> =
@@ -606,7 +604,7 @@ pub(crate) async fn tx_begin_stream(
             if tx.begin {
                 tx_ensure_begin(tx).await?;
             }
-            if opts.commit_tx.unwrap_or(false) {
+            if commit_at_end {
                 tx_before_commit(tx).await?;
             }
             let (mut client, req) =
@@ -777,8 +775,8 @@ pub(crate) async fn tx_rollback(tx: &mut TxExecContext) -> YdbResult<()> {
     active.lease.finish(result)
 }
 
-/// Best-effort rollback when [`super::Transaction`] is dropped without `commit`/`rollback`.
-pub(crate) fn finish_query_tx_on_drop(connection_manager: GrpcConnectionManager, active: ActiveTx) {
+/// Release an unfinished transaction, rolling it back first when its id is known.
+pub(super) fn release_unfinished_tx(connection_manager: GrpcConnectionManager, active: ActiveTx) {
     let ActiveTx {
         lease,
         server_progress,
@@ -940,7 +938,7 @@ mod unit_tests {
     }
 
     #[tokio::test]
-    async fn definitive_dispatched_error_ends_attempt_and_returns_healthy_session() {
+    async fn begin_in_flight_error_discards_session() {
         let pool = SessionPool::new_explicit_bench(SessionPoolSettings::new().with_limit(1));
         let lease = pool.acquire_explicit().await.expect("acquire test session");
         let session_id = lease.session_id().to_string();
@@ -960,16 +958,16 @@ mod unit_tests {
             issues: vec![],
         });
 
-        ctx.finish_dispatched_error(&error)
+        ctx.fail_attempt(&error)
             .expect("rejected operation must finish the transaction");
 
         assert!(matches!(ctx.state, TxState::AttemptFailed(_)));
-        let reused = pool
+        let replacement = pool
             .acquire_explicit()
             .await
-            .expect("healthy session must return to the pool");
-        assert_eq!(reused.session_id(), session_id);
-        reused.return_to_pool();
+            .expect("session with an unconfirmed begin must be replaced");
+        assert_ne!(replacement.session_id(), session_id);
+        replacement.return_to_pool();
     }
 
     #[tokio::test]
@@ -994,7 +992,7 @@ mod unit_tests {
                 issues: vec![],
             });
 
-            ctx.finish_dispatched_error(&error)
+            ctx.fail_attempt(&error)
                 .expect("temporary failure must finish the local transaction attempt");
 
             assert!(matches!(ctx.state, TxState::AttemptFailed(_)));
@@ -1008,7 +1006,7 @@ mod unit_tests {
     }
 
     #[tokio::test]
-    async fn undetermined_dispatched_error_discards_session() {
+    async fn undetermined_query_error_fails_attempt_and_discards_session() {
         let pool = SessionPool::new_explicit_bench(SessionPoolSettings::new().with_limit(1));
         let lease = pool.acquire_explicit().await.expect("acquire test session");
         let session_id = lease.session_id().to_string();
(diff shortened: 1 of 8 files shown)


ydb/src/client_query/exec.rs
```diff
@@ -505,43 +581,56 @@ pub(crate) async fn transaction_begin_stream(
...
         .await;
     if let Err(err) = &result {
+        if matches!(
+            tx.state,
+            TxState::Active(ActiveTransaction {
+                server: ServerTransaction::BeginInFlight,
+                ..
+            })
+        ) {
+            tx.active_mut()?.server = ServerTransaction::NotStarted;
```
rekby: `BeginInFlight` must not be unconditionally demoted here. If the first response is `UNAVAILABLE`, or a timeout/transport failure occurs after dispatch, `invalidates_server_transaction()` deliberately treats the server outcome as ambiguous. This reset leaves the transaction `Active(NotStarted)`; if the callback swallows the error, `transaction_commit` then reports success without sending any commit RPC. Preserve an ambiguous/in-flight terminal state (and discard the lease) unless it is known that the request was never dispatched.

rekby (commented):
The ownership model is clearer, but one stream-open error path still erases an ambiguous in-flight transaction state and can cause a false successful commit.

ydb/src/client_query/exec.rs
```diff
@@ -440,42 +470,79 @@ async fn transaction_execute_request(
...
+    .await;
+
+    match result {
+        Ok(tx_id) => {
+            tx.active_mut()?.server = ServerTransaction::Started(tx_id);
+            Ok(())
+        }
+        Err(err) => {
+            let active = tx.active_mut()?;
+            active.server = ServerTransaction::NotStarted;
```
rekby: `BeginInFlight` cannot always be reset to `NotStarted`. Once `begin_transaction(...).await` is dispatched, a timeout or transport failure may mean the server created the transaction but its ID was lost. If a callback swallows `tx.begin()`'s error, `transaction_commit` then sees `NotStarted` and reports success without sending any finalization RPC—the same false-commit outcome this state machine prevents for lazy `ExecuteQuery`. Distinguish failures known to occur before dispatch; otherwise transition to `Ambiguous` and discard the lease.

ydb/src/client_query/exec.rs
```diff
@@ -505,44 +584,60 @@ pub(crate) async fn transaction_begin_stream(
...
-                owned_lease: None,
-            })
+            let tx_id = stream.take_captured_tx_id();
+            apply_stream_tx_id(tx, tx_id);
+            Ok(stream)
         })
         .await;
     if let Err(err) = &result {
-        transaction_handle_query_error(tx, err);
+        if matches!(
```
rekby: This branch also catches definitive YDB statuses from the first response part, such as `ABORTED` or `BAD_SESSION`, before `transaction_handle_query_error` can classify them as `Invalidated`. If the callback swallows such a first-query error, `resolve_post_callback_action` now fails instead of safely retrying the transaction. Reserve `Ambiguous` for transport/ambiguous statuses; route errors where `invalidates_server_transaction()` is true through the normal invalidation path.

rekby (commented):
The prior first-query false-commit path is fixed, but the new state machine still has an equivalent path for explicit begin and over-classifies definitive first-response errors as ambiguous.

ydb/src/errors.rs
```diff
@@ -297,37 +305,44 @@ impl YdbError {
...
+            Self::YdbStatusError(status) => match status.operation_status() {
+                Ok(StatusCode::Undetermined | StatusCode::Unspecified | StatusCode::Success)
+                | Err(_) => TransactionErrorOutcome::OutcomeUnknown,
+                Ok(
+                    StatusCode::BadRequest
+                    | StatusCode::Unauthorized
+                    | StatusCode::InternalError
+                    | StatusCode::Aborted
+                    | StatusCode::Unavailable
+                    | StatusCode::Overloaded
```
rekby: `UNAVAILABLE` and `OVERLOADED` do not prove that an already-started transaction ended. This module classifies both as `NeedRetry::True`—whose contract says the operation was guaranteed not to complete—and the pre-PR behavior rolled the existing transaction back before retrying. Classifying them as `AttemptFailed` now makes `finish_dispatched_error` set `Invalidated` and return the healthy lease without rollback, so the next attempt can reuse a session that may still host the previous transaction. Preserve a cleanup path that rolls back or discards that lease before retrying; these statuses cannot use the same reusable-session path as `ABORTED`.

rekby (commented):
The new push fixes the two prior state-transition findings, but the new outcome classifier regresses cleanup for transient failures in an already-started transaction.

rekby (commented):
OK

ydb/src/client_query/exec.rs
```diff
@@ -50,55 +49,132 @@ pub(crate) struct ClientExecContext {
...
+pub(crate) struct ActiveTransaction {
+    lease: SessionPoolLease,
+    server: ServerTransaction,
+}
+
+/// Server-side progress within an active transaction.
+///
+/// In-flight states retain the lease in the transaction so cancellation is conservative: dropping
+/// the transaction discards the session instead of issuing a second finalization RPC.
+enum ServerTransaction {
```
rekby: У нас вроде только серверные транзакции и есть. Сейчас в SDK уже есть разные транзакции: Transaction, TxState, ActiveTransaction.

Чем ServerTransaction отличается от TxState? Тут кажется хорошо бы придумать как эти все имена между собой подружить чтобы было понятно чем они друг от друга отличаются.

ydb/src/client_query/exec.rs
```diff
@@ -50,55 +49,132 @@ pub(crate) struct ClientExecContext {
...
+        match &self.state {
+            TxState::Active(ActiveTransaction {
+                server: ServerTransaction::Started(id),
+                ..
+            }) => Some(id),
+            _ => None,
+        }
     }
+
+    fn take_active(&mut self, replacement: TxState) -> YdbResult<ActiveTransaction> {
```
rekby: Это больше похоже на replace, чем на take

ydb/src/client_query/exec.rs
```diff
@@ -50,55 +49,132 @@ pub(crate) struct ClientExecContext {
...
+    pub(crate) session: ClientQuerySession,
+}
+
+pub(crate) enum ClientQuerySession {
+    ServerImplicit,
+    Pooled(SessionPoolLease),
+}
+
+impl ClientQuerySession {
+    pub(crate) fn complete(self) {
```
rekby: Давай этот метод назовём release или reutrn_to_pool, а то кажется что сессию мы завершаем.

ydb/src/errors.rs
```diff
@@ -297,37 +308,45 @@ impl YdbError {
...
-                | StatusCode::Unsupported
-                | StatusCode::SessionBusy
-                | StatusCode::ExternalError
-        )
+    /// Classify the outcome of a dispatched transaction operation independently from retry and
+    /// session-health policy.
+    pub(crate) fn transaction_error_outcome(&self) -> TransactionErrorOutcome {
+        match self {
+            Self::YdbStatusError(status) => match status.operation_status() {
+                Ok(StatusCode::Undetermined | StatusCode::Unspecified | StatusCode::Success)
```
rekby: А как у нас может быть OK с каким-то статусом, кроме Success?

Тут нужно `Ok(Success)` ->  вообще не ошибка, состояние транзакци/сессии не меняется.
по ydb-ошибкам мы сюда попадать не должны.
И как-будто теперь для варианта OK в Enum нет подходящего варианта - оставить транзакцию и сессию как были.

Ok(_) - что-то странное, на вид можно считать как внутренюю неретраибельную ошибку SDK.

а вот дальше - уже делить.

ydb/src/client_query/exec.rs
```diff
@@ -50,55 +49,132 @@ pub(crate) struct ClientExecContext {
...
+    Active(ActiveTransaction),
     /// Real, confirmed commit: either `CommitTransaction` succeeded or `commit_tx` completed.
     Committed,
     /// Rollback path was chosen and the SDK must not report a commit.
     RolledBack,
-    /// The server ended the transaction after a definitive status on a query.
+    /// A definitive operation failure ended the local transaction attempt.
     Invalidated(YdbError),
     /// A commit or rollback RPC returned an error, so the local end attempt was not confirmed.
     Ambiguous(YdbError),
```
rekby: Лучше использовать термин: Undetermined

Ambiguous - это двусмысленность. У нас нет двусмысленности, у нас неопределённость.

rekby (changes requested):

rekby (commented):
OK

rekby (commented):

ydb/src/client_query/exec.rs
```diff
@@ -50,55 +49,134 @@ pub(crate) struct ClientExecContext {
...
+
+    fn fail_attempt(&mut self, error: &YdbError) -> YdbResult<()> {
+        let active = self.replace_active(TxState::AttemptFailed(error.clone()))?;
+
+        if error.requires_session_discard() {
+            // Dropping a lease without returning it schedules session cleanup.
+            return Ok(());
+        }
+
+        release_unfinished_tx(self.connection_manager.clone(), active);
```
rekby, line 175: `fail_attempt` now sends ordinary query failures on an established transaction through `release_unfinished_tx`, but that detached cleanup holds the `SessionPoolLease` while awaiting both connection acquisition and `RollbackTransaction` without any timeout. A server that never completes the rollback can therefore retain the pool permit indefinitely; repeated retryable failures can exhaust the session pool even after the enclosing `retry_tx` deadline expires. Please bound this background rollback and discard the lease when that bound is reached, as the synchronous rollback path does with `resolve_effective_timeout`.

rekby (commented):
OK

rekby (approved):

ydb/src/client_query/stream_facade.rs
```diff
@@ -4,67 +4,144 @@ use std::time::Duration;
...
         let dropped_mid_stream = self.stream.in_progress();
         self.stream.cancel();
-        if let ExecCoreRef::Transaction(ctx) = &mut self.core
-            && let Some(lease) = &mut ctx.pooled_lease
+        let lifecycle = std::mem::replace(&mut self.lifecycle, QueryStreamLifecycle::Finished);
+        if let QueryStreamLifecycle::Active(QueryStreamOwner::Tx { context, .. }) = lifecycle
             && dropped_mid_stream
         {
-            lease.invalidate();
+            tx_invalidate_session(context);
```
rekby: Dropping a partially consumed transaction stream only invalidates the lease; it leaves `TxState::Active(Started(_))`. When the callback returns, `resolve_post_callback_action` therefore selects `Commit`, and the new `tx_commit` path no longer calls `ensure_healthy()`, so it can still send `CommitTransaction` on this session. This violates the documented guarantee that dropping the stream early does not commit and may finalize a query whose outcome is unknown. Please transition the transaction to a non-active failure/undetermined state while discarding the lease here, so post-callback processing cannot commit it.

rekby (commented):

rekby (commented):
OK

rekby (commented):
The prior transaction-state findings appear fixed. One review-level issue remains: the description marks this as “no functional changes” and only covers query/session ownership, while the diff also changes false-commit behavior, maps TopicWriterTx `NOT_FOUND` errors to `BAD_SESSION`, and redesigns SLO shutdown verification to commit the final topic event. Please update the description/checklist to explain these behavioral changes and their rationale, or split unrelated changes. Repository checks could not run because `cargo` is unavailable in the review environment.

ydb/src/client_query/mod.rs
```diff
@@ -289,9 +287,15 @@ impl QueryClient {
             .retry(closure!(
                 [&client = self, callback, &options],
                 async |retry: &RetryState| {
+                    let lease = match client.ctx.session_pool.acquire_explicit().await {
+                        Ok(lease) => lease,
+                        Err(err) => {
+                            return YdbOrCustomerError::from(err).retry_flow(idempotency);
```
rekby, line 293: Session acquisition happens before the callback, so no user or transaction operation can have completed yet. Passing the transaction's `idempotency` makes the default non-idempotent `retry_tx` treat transient acquire/create/attach failures (`Transport` is `IdempotentOnly`) as terminal instead of retrying this safe pre-attempt step.

```suggestion
                            return YdbOrCustomerError::from(err).retry_flow(Idempotency::Idempotent);
```

ydb/src/client_query/exec.rs
```diff
@@ -445,265 +489,394 @@ async fn transaction_execute_request(
...
+
+    let terminal = match &result {
         Ok(()) => TxState::Committed,
-        Err(err) => TxState::Ambiguous(err.clone()),
+        Err(err) => TxState::Undetermined(err.clone()),
     };
+    let active = tx.replace_active(terminal)?;
     // Do not retry commit: a transport timeout may mean the commit succeeded server-side.
-    result
+    active.lease.finish(result)
```
rekby, line 720: `SessionPoolLease::finish` returns the lease for errors that do not require session recreation, including YDB `UNAVAILABLE`, `OVERLOADED`, and `UNDETERMINED`. Here the state was just changed to `TxState::Undetermined`, so finalization may not have completed and the session may still host the transaction. Returning it lets another attempt reuse it without cleanup. Discard the lease on every finalization error; the identical `tx_rollback` tail below needs the same treatment.

ydb/src/client_query/stream_facade.rs
```diff
@@ -4,67 +4,141 @@ use std::time::Duration;
...
         };
-        let Some((raw, tx_id)) = next else {
-            return Ok(None);
-        };
-        if let ExecCoreRef::Transaction(ctx) = &mut self.core {
-            apply_stream_tx_id(ctx, tx_id);
+        match next {
+            Some((raw, transaction_id)) => {
+                self.apply_transaction_id(transaction_id);
+                ResultSet::try_from(raw).map(Some)
```
rekby, line 135: A `ResultSet::try_from` failure bypasses `handle_error`, leaving the transaction stream lifecycle active. If the caller handles that conversion error, finishes draining, and calls `close`, `tx_finish_query` (or the post-callback commit) can still commit the transaction despite the reported query failure. Route conversion failures through the transaction error/abort path just like raw stream errors.

rekby (commented):

PR #621, opened 2026-08-14
refactor(query): model dispatched transaction operations

aral01:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no api changes)
- [ ] Build related changes

- Transaction does not track ExecuteQuery RPC in progress. Started transaction still looks ready while query is running.
- On cancellation or Drop SDK can try rollback while another operation is still in progress.
- Session can be reused after operation with unknown server result.
- Transaction session and hooks are finished in different places.

Issue Number: N/A

## What is the new behavior?

- Transaction server state explicitly describes NotStarted, Ready and InFlight states.
- InFlight state describes Begin, Query, Commit and Rollback operations.
- Successful query returns transaction to Ready state.
- Definitive server error invalidates transaction.
- Cancellation or unknown server result makes transaction Ambiguous and discards its session.
- Drop tries rollback only for Ready transaction. InFlight transaction does not send competing rollback.
- Transaction owns its session and hooks until one terminal transition. Hooks are called exactly once.

## Other information

- This prepares transaction cleanup for CleanupWorker. Worker is not introduced in this PR.

### DIFF
--- a/ydb/src/client_query/exec.rs
+++ b/ydb/src/client_query/exec.rs
@@ -80,8 +80,8 @@ pub(crate) enum TxState {
     Committed,
     /// Rollback path was chosen and the SDK must not report a commit.
     RolledBack,
-    /// A conclusive failure ended the local transaction attempt. Its session ownership has already
-    /// been resolved; retry policy decides whether to start another attempt.
+    /// A returned operation error ended the local transaction attempt. Its session ownership has
+    /// already been resolved; retry policy decides whether to start another attempt.
     AttemptFailed(YdbError),
     /// A dispatched operation's server outcome was not conclusively observed.
     Undetermined(YdbError),
@@ -243,9 +243,9 @@ impl TxExecContext {
         replacement: TxState,
         status: QueryTxCommitStatus,
     ) -> YdbResult<SessionPoolLease> {
-        let mut active = self.replace_active(replacement)?;
-        active.notify_hooks(status);
-        Ok(active.lease)
+        let mut previous_active = self.replace_active(replacement)?;
+        previous_active.notify_hooks(status);
+        Ok(previous_active.lease)
     }
 
     /// End an attempt with an unknown server outcome and discard its session.
@@ -296,17 +296,6 @@ impl TxExecContext {
         Ok(())
     }
 
-    fn resolve_finalization_error(&mut self, error: &YdbError) -> YdbResult<()> {
-        // An error that is safe to retry even for non-idempotent work guarantees that the
-        // finalization did not complete. Other retry classes do not prove the outcome of an
-        // already-dispatched commit or rollback, so keep those outcomes undetermined.
-        if error.is_retriable(Idempotency::NonIdempotent) {
-            self.fail_attempt(error)
-        } else {
-            self.finish_undetermined(error.clone())
-        }
-    }
-
     fn handle_pre_dispatch_error(&mut self, error: &YdbError) -> YdbResult<()> {
         if error.requires_session_discard() {
             self.fail_attempt(error)?;
@@ -355,14 +344,14 @@ impl Drop for TxExecContext {
             (TxOperationState::InFlight, _) => return,
         };
 
+        let cleanup_timeout = lease.cleanup_timeout();
+        let session_id = lease.session_id().to_string();
+
         spawn_pool_release(async move {
-            let rollback_ok = client
-                .rollback_transaction(lease.session_id(), tx_id.as_str())
-                .await
-                .is_ok();
-            if rollback_ok {
-                lease.return_to_pool();
-            }
+            let rollback = client
+                .rollback_transaction(&session_id, tx_id.as_str())
+                .map_err(YdbError::from);
+            finish_rollback_cleanup(lease, cleanup_timeout, rollback).await;
         });
     }
 }
@@ -855,7 +844,7 @@ pub(crate) async fn tx_commit(tx: &mut TxExecContext) -> YdbResult<()> {
             Ok(())
         }
         Err(error) => {
-            tx.resolve_finalization_error(&error)?;
+            tx.fail_attempt(&error)?;
             Err(error)
         }
     }
@@ -899,7 +888,7 @@ pub(crate) async fn tx_rollback(tx: &mut TxExecContext) -> YdbResult<()> {
             Ok(())
         }
         Err(error) => {
-            tx.resolve_finalization_error(&error)?;
+            tx.fail_attempt(&error)?;
             Err(error)
         }
     }
@@ -1080,31 +1069,6 @@ mod unit_tests {
             assert_ne!(replacement.session_id(), session_id);
             replacement.return_to_pool();
         }
-    }
-
-    #[tokio::test]
-    async fn undetermined_finalization_discards_session() {
-        let pool = SessionPool::new_explicit_bench(SessionPoolSettings::new().with_limit(1));
-        let lease = pool.acquire_explicit().await.expect("acquire test session");
-        let session_id = lease.session_id().to_string();
-        let mut ctx = test_transaction_context(lease).await;
-        ctx.mark_query_in_flight_for_test("tx-1");
-        let error = YdbError::YdbStatusError(crate::errors::YdbStatusError::new(
-            "transaction outcome unknown",
-            StatusCode::Undetermined as i32,
-            vec![],
-        ));
-
-        ctx.resolve_finalization_error(&error)
-            .expect("unknown outcome must finish the transaction");
-
-        assert!(matches!(ctx.state, TxState::Undetermined(_)));
-        let replacement = pool
-            .acquire_explicit()
-            .await
-            .expect("discarded session must be replaced");
-        assert_ne!(replacement.session_id(), session_id);
-        replacement.return_to_pool();
     }
 
     #[tokio::test]
--- a/ydb/src/client_query/exec.rs
+++ b/ydb/src/client_query/exec.rs
@@ -1045,7 +1045,7 @@ mod unit_tests {
     }
 
     #[tokio::test]
-    async fn transient_dispatched_error_discards_session_with_possibly_active_transaction() {
+    async fn transient_dispatched_error_discards_session_when_cleanup_fails() {
         for status in [StatusCode::Unavailable, StatusCode::Overloaded] {
             let pool = SessionPool::new_explicit_bench(SessionPoolSettings::new().with_limit(1));
             let lease = pool.acquire_explicit().await.expect("acquire test session");
--- a/ydb/src/client_query/exec.rs
+++ b/ydb/src/client_query/exec.rs
@@ -5,8 +5,8 @@ use std::time::{Duration, Instant};
 use futures_util::TryFutureExt;
 use tokio::time::timeout;
 
-use crate::errors::{Idempotency, YdbError, YdbResult};
 use crate::client_metrics::names::MetricsNames;
+use crate::errors::{Idempotency, YdbError, YdbResult};
 use crate::grpc_connection_manager::GrpcConnectionManager;
 use crate::grpc_wrapper::raw_query_service::client::RawQueryClient;
 use crate::grpc_wrapper::raw_query_service::execute_query::RawExecuteQueryRequest;
@@ -823,6 +823,7 @@ pub(crate) async fn tx_commit(tx: &mut TxExecContext) -> YdbResult<()> {
         return Err(err);
     }
     let operation_timeout = remaining_retry_timeout(tx.retry_deadline);
+    let commit_counter = tx.metrics_names.client_transaction_commit_counter.clone();
     let active = tx.active_mut()?;
     if active.server_progress.operation.is_in_flight() {
         return Err(TxServerProgress::operation_in_progress_error());
@@ -833,9 +834,7 @@ pub(crate) async fn tx_commit(tx: &mut TxExecContext) -> YdbResult<()> {
         return Ok(());
     };
 
-    tx.metrics_names
-        .client_transaction_commit_counter
-        .increment(1);
+    commit_counter.increment(1);
 
     debug_assert_eq!(active.server_progress.operation, TxOperationState::Ready);
     active.server_progress.operation = TxOperationState::InFlight;
@@ -1006,7 +1005,13 @@ mod unit_tests {
             .get_auth_service_to_node(RawQueryClient::new, lease.node_uri())
             .await
             .expect("create test query client");
-        tx_exec_context(client, lease, TransactionOptions::default(), None)
+        tx_exec_context(
+            client,
+            lease,
+            TransactionOptions::default(),
+            None,
+            MetricsNames::new(None),
+        )
     }
 
     #[test]
--- a/ydb/src/client_query/exec.rs
+++ b/ydb/src/client_query/exec.rs
@@ -2,9 +2,11 @@ use std::collections::HashMap;
 use std::future::Future;
 use std::time::{Duration, Instant};
 
+use futures_util::TryFutureExt;
 use tokio::time::timeout;
 
 use crate::errors::{Idempotency, YdbError, YdbResult};
+use crate::client_metrics::names::MetricsNames;
 use crate::grpc_connection_manager::GrpcConnectionManager;
 use crate::grpc_wrapper::raw_query_service::client::RawQueryClient;
 use crate::grpc_wrapper::raw_query_service::execute_query::RawExecuteQueryRequest;
@@ -15,13 +17,13 @@ use crate::grpc_wrapper::raw_query_service::transaction_control::{
 use crate::retry_settings::RetrySettings;
 use crate::traces::helpers::ensure_len_string;
 
-use crate::client_metrics::names::MetricsNames;
-use crate::session_pool::{SessionPool, SessionPoolLease, spawn_pool_release};
 use crate::types::Value;
 use crate::{TransactionOptions, TxMode, closure};
 use tracing::instrument;
 
-use super::hooks::QueryTxHook;
+use crate::session_pool::{SessionPool, SessionPoolLease, spawn_pool_release};
+
+use super::hooks::{QueryTxCommitStatus, QueryTxHook};
 
 #[derive(Clone, Debug, Default)]
 pub(crate) struct CallOptions {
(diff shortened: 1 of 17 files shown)


ydb/src/client_query/exec.rs
```diff
@@ -154,10 +318,52 @@ impl TransactionExecContext {
...
+    fn finish_query_error(&mut self, error: &YdbError) -> YdbResult<()> {
+        if error.invalidates_server_transaction() {
+            return self.finish_server_rejected(error);
+        }
+        if error.requires_session_discard() {
+            return self.finish_ambiguous(error.clone());
+        }
+        match self.active_mut()?.server.finish_failed_query()? {
+            FailedQueryState::Ready => Ok(()),
+            FailedQueryState::OutcomeUnknown => self.finish_ambiguous(error.clone()),
```
rekby: `UNDETERMINED` falls through this logic: it neither invalidates the server transaction nor requires session discard, so an existing transaction is restored to `Ready` even though the query outcome is unknown. The caller can then dispatch another operation on the same transaction, contrary to the stated ambiguous-outcome behavior. Classify unknown operation outcomes separately from session health and transition them through `finish_ambiguous`.

ydb/src/client_query/exec.rs
```diff
@@ -472,49 +674,40 @@ async fn transaction_execute_request(
...
+            tx.active_mut()?.server = ServerTransaction::Ready(tx_id);
             Ok(())
         }
         Err(err) => {
-            let active = tx.active_mut()?;
-            active.server = ServerTransaction::NotStarted;
-            if err.requires_session_discard() {
-                active.lease.invalidate();
-            }
+            tx.finish_ambiguous(err.clone())?;
```
rekby: This treats every `BeginTransaction` failure as ambiguous. Definitive server statuses such as `ABORTED` satisfy `invalidates_server_transaction()` and may be retryable without requiring session discard; marking them `Ambiguous` deletes the lease and, if the callback catches the error, makes `resolve_post_callback_action` fail instead of retry. Route definitive responses through `finish_server_rejected` (or equivalent), reserving this path for unknown outcomes.

ydb/src/client_query/exec.rs
```diff
@@ -696,44 +866,36 @@ pub(crate) async fn transaction_commit(tx: &mut TransactionExecContext) -> YdbRe
...
-    // Do not retry commit: a transport timeout may mean the commit succeeded server-side.
-    active.lease.finish(result)
+    match result {
+        Ok(()) => {
+            tx.finish_transaction(TxState::Committed, QueryTxCommitStatus::Committed)?
+                .return_to_pool();
+            Ok(())
+        }
+        Err(error) => {
+            tx.finish_ambiguous(error.clone())?;
```
rekby: All commit failures now drop the session via `finish_ambiguous`, including definitive YDB statuses for which `requires_session_discard()` is false. The previous `lease.finish(result)` returned those healthy sessions to the pool, so repeated server-side rejections now cause unnecessary DeleteSession/CreateSession churn. Classify definitive responses separately; the rollback error branch has the same issue.

rekby (commented):

rekby (commented):
OK

ydb/src/client_query/stream_facade.rs
```diff
@@ -46,15 +46,21 @@ impl Drop for QueryStream<'_> {
...
+    }
+
+    fn finish_cancelled_lifecycle(&mut self, dropped_mid_stream: bool) {
         let lifecycle = std::mem::replace(&mut self.lifecycle, QueryStreamLifecycle::Finished);
         if let QueryStreamLifecycle::Active(QueryStreamOwner::Transaction { context, .. }) =
             lifecycle
             && dropped_mid_stream
         {
-            transaction_invalidate_session(context);
+            transaction_cancel_query(context);
```
rekby: This terminal transition is skipped when the response has already reached EOF because `dropped_mid_stream` is then false. However, `next_result_set()` returning `None` does not call `transaction_finish_query`, so dropping a fully drained but unclosed stream leaves the transaction in `InFlight(Query(_))`; after the borrow is released, every subsequent query, commit, or rollback fails with “operation is already in progress.” A conversion error on the final result set reaches the same state. Make every unclosed transaction stream terminal on drop, even after EOF, or explicitly finish the transaction when EOF is observed.

ydb/src/client_query/exec.rs
```diff
@@ -696,44 +835,36 @@ pub(crate) async fn transaction_commit(tx: &mut TransactionExecContext) -> YdbRe
...
-    // Do not retry commit: a transport timeout may mean the commit succeeded server-side.
-    active.lease.finish(result)
+    match result {
+        Ok(()) => {
+            tx.finish_transaction(TxState::Committed, QueryTxCommitStatus::Committed)?
+                .return_to_pool();
+            Ok(())
+        }
+        Err(error) => {
+            tx.finish_dispatched_error(&error)?;
```
rekby: Status-coded commit failures are now classified as `Invalidated`, but `try_attempt_body` still wraps every `tx.commit()` error in `ControlFlow::Break`, bypassing `retry_flow`. Consequently, `retry_tx` will not retry a definitive `ABORTED`/`UNAVAILABLE` commit failure even though these statuses are `NeedRetry::True` and this code establishes that the attempt did not commit. The caller should continue through the retry policy for `Invalidated`, while retaining the unconditional break for `Ambiguous` commit outcomes.

rekby (commented):

rekby (commented):
OK

ydb/src/client_query/exec.rs
```diff
@@ -532,44 +696,58 @@ pub(crate) fn transaction_finish_query(
...
-        tracing::error!(%error, "failed to finish transaction after query error");
+pub(crate) fn transaction_handle_query_error(
+    tx: &mut TransactionExecContext,
+    err: &YdbError,
+) -> YdbResult<()> {
+    if tx.active()?.server.query_is_in_flight() {
+        return tx.finish_dispatched_error(err);
     }
+    if err.requires_session_discard() {
+        return tx.finish_ambiguous(err.clone());
```
rekby: This path also handles failures before `start_query()` (for example, `ensure_healthy()` returning `BAD_SESSION`). No transaction operation was dispatched, so marking the state `Ambiguous` makes `retry_tx` stop even though `BAD_SESSION` is classified as retryable and the attempt definitely did not commit. Terminate it as a failed attempt while still discarding the unhealthy session.

```suggestion
        return tx.finish_attempt_failed(err);
```

rekby (commented):

rekby (commented):
OK

rekby (commented):
OK

rekby (commented):
OK

ydb/src/errors.rs
```diff
@@ -388,7 +392,10 @@ impl YdbError {
             }
             Self::YdbStatusError(status) => match StatusCode::try_from(status.operation_status) {
                 Ok(
-                    StatusCode::BadSession | StatusCode::SessionBusy | StatusCode::SessionExpired,
+                    StatusCode::BadSession
+                    | StatusCode::NotFound
```
rekby: `NOT_FOUND` is now treated as evidence that the pooled session vanished, but `need_retry()` still classifies it as `NeedRetry::False`. Consequently, `SessionPoolLease::finish` discards the lease, yet `client_begin_stream`/`retry_tx` immediately return the error instead of acquiring a replacement and retrying the same operation. If Query Service `NOT_FOUND` means a missing session as this change and its test assume, it must be retriable like `BAD_SESSION`; if it can also describe another missing resource, this generic classifier lacks enough context to discard it globally. Please classify the contextual missing-session case consistently and test recovery of the failing operation, rather than only the next lease acquisition.

rekby (commented):

ydb/src/client_query/exec.rs
```diff
@@ -162,73 +230,173 @@ impl TxExecContext {
...
+
+        spawn_pool_release(async move {
+            let rollback_ok = client
+                .rollback_transaction(lease.session_id(), tx_id.as_str())
+                .await
+                .is_ok();
+            if rollback_ok {
+                lease.return_to_pool();
+            }
+        });
```
rekby, line 355: This new drop cleanup no longer applies the lease's cleanup timeout. If rollback stalls, the spawned future retains the session-pool permit indefinitely, so repeated dropped transactions can exhaust the pool. Route it through the same bounded helper used by `fail_attempt`.

```suggestion
        let cleanup_timeout = lease.cleanup_timeout();
        let session_id = lease.session_id().to_string();
        spawn_pool_release(async move {
            let rollback = client
                .rollback_transaction(&session_id, tx_id.as_str())
                .map_err(YdbError::from);
            finish_rollback_cleanup(lease, cleanup_timeout, rollback).await;
        });
```

rekby (commented):

ydb/src/client_query/exec.rs
rekby: This is only partially fixed: `resolve_finalization_error` uses `is_retriable(NonIdempotent)` as the definitive/ambiguous discriminator. Consequently, received `BAD_REQUEST` or `UNAUTHORIZED` commit/rollback responses still go through `finish_undetermined` and discard a healthy lease, even though these are conclusive server rejections with `requires_session_discard() == false`. Please distinguish observed YDB status responses from transport/timeout uncertainty.

ydb/src/client_query/exec.rs
```diff
@@ -162,73 +230,173 @@ impl TxExecContext {
...
+        self.active_mut()?.hooks.push(hook);
+        Ok(())
+    }
+
+    fn finish_tx(
+        &mut self,
+        replacement: TxState,
+        status: QueryTxCommitStatus,
+    ) -> YdbResult<SessionPoolLease> {
+        let mut active = self.replace_active(replacement)?;
```
rekby: тут скорее previous, чем active. т.к. активное состояние - replacement.

ydb/src/client_query/query_hooks_integration_test.rs
```diff
@@ -83,7 +83,8 @@ impl QueryTxHook for StatsHook {
 }
 
 fn register_stats_hook(tx: &mut Transaction, stats: &Arc<HookStats>) {
-    tx.register_hook(stats.hook());
+    tx.register_hook(stats.hook())
```
rekby, line 86: expect - это потенциальная паника. Тут причин для паники нет - register_hook исходя из кода всегда возвращает Ok, т.е. можно убрать Result-обёртку из кода и не полагаться на runtime-проверки и паники.

rekby (commented):
OK

ydb/src/client_topic/topicreader/reader_tx.rs
```diff
@@ -62,7 +62,7 @@ impl<'a> TopicReaderTx<'a> {
 
         tx.register_hook(ReaderTxHook {
             runtime: runtime.clone(),
-        });
+        })?;
```
rekby, line 65: в текущей реализации `register_hook` всегда возвращает ок. Менять его тип и код вокруг вызовов не нужно - проще и надёжнее просто убрать Result из возвращаемого типа.

rekby (approved):
Нужно поправить `.expect` и можно мерджить

rekby (commented):
OK

ydb/src/errors.rs
rekby: This remains unresolved: `fail_attempt` treats every YDB `NOT_FOUND` as `session_was_not_found` and discards the lease, while `need_retry()` still returns `False`. The new `finalization_not_found_discards_session_without_retry` test explicitly locks in failing the current operation and only recovering on a later call. Please identify missing-session responses contextually and make that case retryable; otherwise unrelated `NOT_FOUND` responses are also misclassified as session loss.

ydb/src/client_query/exec.rs
rekby: The retry-state part is fixed, but lease handling is still unresolved: `fail_attempt` takes the `(InFlight, None)` arm and returns, dropping the lease. This is the explicit/lazy-begin state before a transaction ID is received; definitive statuses such as `ABORTED` or `BAD_REQUEST` have `requires_session_discard() == false`, so the healthy session should be returned instead of deleted.

ydb/src/client_query/mod.rs
```diff
@@ -230,41 +232,47 @@ impl QueryClient {
...
-                PostCallbackAction::Retry(err) => {
-                    tx.notify_hooks(QueryTxCommitStatus::Aborted);
-                    ControlFlow::Continue(err.into())
+                let mut retry_error = retry_error_before_cleanup(&tx.ctx.state, err);
+                if let Err(error) = tx_rollback(&mut tx.ctx).await {
+                    tracing::warn!(
+                        %error,
+                        "rollback after transaction callback failure did not complete"
+                    );
+                    retry_error = retry_error_for_operation(&tx.ctx.state, error);
```
rekby: This can make a customer error retryable. When the callback returns `YdbOrCustomerError::Customer` while the transaction is active, `retry_error_before_cleanup` preserves that non-retriable error; however, a retryable rollback failure such as `BAD_SESSION` replaces it here, and `retry_flow` then executes the entire callback again—even for the default non-idempotent callback because `BAD_SESSION` is `NeedRetry::True`. Cleanup failure must not turn an application-level failure into permission to repeat arbitrary callback side effects. Preserve the customer error as the retry cause (while still recording/returning an undetermined cleanup outcome as needed), and add a regression test with a customer error followed by a retryable rollback failure.

rekby (commented):

ydb/src/client_query/exec.rs
rekby: This remains in the new state machine: `fail_attempt` still takes `(TxOperationState::InFlight, None) => return Ok(())`, which drops the destructured lease. A definitive explicit/lazy begin rejection with `requires_session_discard() == false` reaches this arm, so the healthy session is still deleted instead of returned.

ydb/src/client_query/exec.rs
rekby: Lease reuse is still only partially addressed. `finish_rollback_cleanup` returns the lease solely for `Ok(Ok(()))`; a conclusive rollback response such as `NOT_FOUND` or `BAD_REQUEST` drops it even though `requires_session_discard() == false`. Thus cleanup after definitive transaction failures can still delete healthy sessions.

ydb/src/client_query/exec.rs
rekby: This remains in the rewritten state machine. `tx_ensure_begin` sets `operation = InFlight` before awaiting, and `fail_attempt` still takes `(InFlight, None) => return Ok(())`; the destructured lease is then dropped. Thus a conclusive `ABORTED`/`BAD_REQUEST` begin response still deletes a healthy session, and the new `tx_id = None` test explicitly locks in that churn.

ydb/src/client_query/exec.rs
rekby: Still unresolved: `finish_rollback_cleanup` returns the lease only for `Ok(Ok(()))`. Any conclusive rollback status for which `requires_session_discard()` is false—particularly `NOT_FOUND` after the original failure already ended the transaction—drops the healthy session. The cleanup should apply the existing session-discard classifier before deciding whether to reuse the lease.

sshaplygin:
@rekby @aral01 `definitive_commit_failure_retries_whole_transaction`, added here in `b7696b5`, is racy. It failed the `tests (RUST_VERSION_MODERN)` leg of #588, a PR that does not touch this code. Details below, plus a deterministic way to reproduce it.

## The failure

```
---- definitive_commit_failure_retries_whole_transaction stdout ----
thread 'definitive_commit_failure_retries_whole_transaction' panicked at ydb/tests/query_tx.rs:435:5:
assertion `left == right` failed: successful rollback cleanup must return the session for retry
  left: 2
 right: 1
```

In that run the MSRV (1.88) and the stable/latest-deps legs passed; only 1.96 failed.

## Why it races

Cleanup after a failed commit is a detached task — `ydb/src/client_query/exec.rs`:

```rust
spawn_pool_release(async move {
    let rollback = client.rollback_transaction(&session_id, tx_id.as_str()).map_err(YdbError::from);
    finish_rollback_cleanup(lease, cleanup_timeout, rollback).await;
});
```

and the lease goes back only after the rollback RPC has answered:

```rust
if matches!(timeout(cleanup_timeout, rollback).await, Ok(Ok(()))) {
    lease.return_to_pool();
}
```

`retry_tx` does not wait for that task — the next attempt asks the pool for a session straight away. Two outcomes, both legal:

- the detached task returned the lease first → the attempt reuses the session → `create_session_count == 1`;
- the attempt got there first → the pool is empty → `CreateSession` → **2**.

`wait_for_rollback_count(&tx_lifecycle, 1)` does not order this. It runs after `retry_tx` has already returned, so the second attempt has long since taken or created its session; it only observes that the mock received the rollback. The test uses `#[tokio::test]`, i.e. the current-thread runtime, so the outcome comes down to which ready task the scheduler polls first.

## Reproducing it

Deterministically, on master, by delaying the detached task in `handle_commit_error`:

```rust
spawn_pool_release(async move {
    tokio::time::sleep(Duration::from_millis(50)).await; // force the retry to win
    let rollback = client.rollback_transaction(&session_id, tx_id.as_str())
```

With that delay the test fails on every run with exactly `left: 2, right: 1`.

Without it, the failure needs a slow enough machine. It did not reproduce here in 330 local runs — 320 of the single test (8 concurrent workers), 25 of the whole `query_tx` binary, and 10 under `cargo llvm-cov` with the same instrumentation CI uses.

## Where it surfaced

The diff of #588 under `ydb/src` is six files: the `tracing`/`rand` imports in `client_coordination`, `client_table_test_integration.rs` (`#[ignore]` tests), the two load balancers and `retry_settings.rs` — all `rand` 0.8 → 0.9 call-site renames. Nothing in `client_query`, the session pool or the mock server. It does bump `tokio` 1.38 → 1.53, which is the kind of thing that shifts scheduling, but the ordering the assertion relies on was never guaranteed to begin with.

Master has had exactly one CI run since this merged, so a green master is a single observation rather than evidence that the test is stable.

## Suggested fix

Give that one test a client whose session pool holds a single session. The second attempt then has no permit to take and has to wait for the cleanup to hand the session back, which is exactly the behaviour the assertion is about — and it holds regardless of scheduling:

```rust
async fn make_client_with_single_session(server: &MockServer) -> YdbResult<Client> {
    make_client(server)
        .await?
        .with_session_pool(SessionPoolSettings::new().with_limit(1))
        .await
}
```

Checked with the 50 ms delay above still in place: the test passes. The whole `query_tx` suite stays green (23/23).

I have this on a branch off master and can open it as a PR, or leave it to you if a follow-up to this one is already planned.

sshaplygin:
Fix opened as #637.

PR #626, opened 2026-08-17
feat(query)!: stream logical result sets lazily

### LINKED ISSUE #617: feat(query): add partial Result Set streaming
# Feature Request

**Describe the Feature Request**

Currently, QueryStreaming stream unit is whole ResultSet, which is accumulated in memory. 

Let QueryStream stream parts of Result Sets, as grpc sends them, to allow reduce memory consumption and time between parts processing.

**Describe Preferred Solution**

**Describe Alternatives**

**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**

- [ ] Yes
- [ ] No

_(Help can be provided if you need assistance submitting a PR)_

**Potential Impact**

### LINKED ISSUE #618: bug: materialize query_client.query result
# Bug Report

query_client.query method return real grpc_stream instead of memoized results, similary to nearby methods, for example `query_result_set`.

**Expected behavior:**

1. All methods on query_client return fully in-memory result. Streaming allow in retry_tx method for interactive transactions.
2. All query_client methods has internal retries, similar to exec, query_result_set, query_row

aral01:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Query provides ResultSet streaming API. Each result set is accumulated in memory. Transport errors must be retried manually.

Issue Number: closes #618, closes #617

## What is the new behavior?

- Query without transaction provides only materialized API: method returns all ResultSets. Errors are retried internally. 

- Transaction provides partial result set streaming API, where each result set is lazily streamed.


DXist:
@aral01 If we materialize the results of the `query` method for a one-shot query, do we plan a streamed alternative for this type of query? Why do we require an interactive transaction for a streamed one-shot query?

Streamed one-shot allows a custom retry strategy for large results. For example, retry from a not yet processed starting point instead of from the same start as the first query attempt.

PR #634, opened 2026-08-26
fix(slo): topic tx slo reliability

aral01:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- Fix Prometheus `rate(counter[5s])` calculation by initializating initial error counts as 0
- Change final TopicTx validation. Use one transaction with commit to ensure topic/table consistency. Compare table and topic output.

### DIFF
--- a/tests/slo/native/topic-tx/src/storage/verification.rs
+++ b/tests/slo/native/topic-tx/src/storage/verification.rs
@@ -2,50 +2,41 @@ use std::time::Duration;
 
 use anyhow::{Context, Result, ensure};
 use tokio::time::{Instant, sleep};
-use ydb::{DescribeConsumerOptionsBuilder, Transaction, YdbOrCustomerError, closure};
+use ydb::{Transaction, YdbOrCustomerError, closure};
 
-use slo_framework::topic_tx::{ChainTransition, PartitionId, TopicOffset};
+use slo_framework::topic_tx::{ChainTransition, PartitionId};
 
 use super::TopicTxStorage;
 use super::queries::{read_next_transition, required_field};
 use super::transaction::{invalid_chain_state, reader_options};
 
 const POOL_RELEASE_TIMEOUT: Duration = Duration::from_secs(5);
-const POOL_POLL_INTERVAL: Duration = Duration::from_millis(10);
+const STATE_POLL_INTERVAL: Duration = Duration::from_millis(10);
 
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub(super) struct PartitionOffsets {
-    pub(super) partition_id: PartitionId,
-    pub(super) committed_offset: TopicOffset,
-    pub(super) end_offset: TopicOffset,
+struct PartitionObservation {
+    partition_id: PartitionId,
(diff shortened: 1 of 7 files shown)


tests/slo/native/topic-tx/src/storage/verification.rs
```diff
@@ -57,68 +48,34 @@ impl TopicTxStorage {
...
-                    // roll it back so verification does not advance the consumer.
                     let transition =
                         read_next_transition(reader, tx, *partition_id, *deadline).await?;
-                    verify_partition_table_state(tx, table_path, &transition).await?;
-                    tx.rollback().await?;
-                    Ok(())
+                    read_partition_table_observation(tx, table_path, &transition).await
                 }
             ))
             .idempotent(true)
```
rekby, line 61: [P1] This does not retry a serialization conflict reported by the implicit commit. `QueryClient::try_attempt_body` returns `ControlFlow::Break` for every `tx.commit()` error, so `.idempotent(true)` only permits retries for errors raised before that commit. If a late worker commit wins the race, verification can receive `ABORTED` from `CommitTransaction` and fail instead of reconnecting and observing the committed state, contrary to the new reliability model and README. Please explicitly retry confirmed-aborted commit outcomes while still treating transport/ambiguous outcomes as terminal, and cover this race with a test.

rekby (commented):

tests/slo/slo-framework/src/metrics.rs
```diff
@@ -142,6 +142,19 @@ impl Metrics {
         record_latency_series(&self.inner.topic_e2e_latency, latency, attrs_key);
     }
 
+    pub fn initialize_error_series(&self, operation_type: OperationType, error_name: &'static str) {
```
rekby, line 157: [P1] Seeding only `sdk.errors.total` leaves the same first-sample loss in the failure-labeled counters used by the other TopicTx SLOs. `Span::finish(Some(_), ...)` creates `sdk.operations.total{operation_status="failure"}` and `sdk.retry.attempts.total{operation_status="failure"}` at nonzero values; Prometheus `rate`/`increase` cannot recover the increment preceding that label set's first sample. Consequently, a single failure can still leave `transaction_availability` at 100%, and its retry attempts can be omitted from `transaction_retry_overhead_percent`. Please initialize the corresponding failure operation/retry counter series at zero before workers start as well.

tests/slo/native/topic-tx/src/storage/verification.rs
rekby: This remains unresolved after the new push; the delta only adds operation/retry metric initialization. In `QueryClient::try_attempt_body`, an implicit `tx.commit()` error is wrapped in `ControlFlow::Break`, and `?` returns it before `.retry_flow(idempotency)` is reached (`ydb/src/client_query/mod.rs:242–265`). Thus a confirmed `ABORTED` commit still terminates verification instead of retrying, while the README now promises that conflict will be retried.

PR #641, opened 2026-09-21
fix sessions default timeouts, its flacked for slow VMs 

rekby:
For example x86 on ARM CPU.


rekby (commented):
OK

